PR #19, opened 2024-07-18
Set up Rails

tonyheadford:
This Rails app was generated using DfE's Rails template. We have customised it in the following ways:

* removed many of the Rails-genreated comments in config files
* added some common inflections like API and ECT
* added FactoryBot
* use NPM instead of Yarn (this means calling esbuild and sass directly)
* bumped to Ruby 3.3.4
* bumped to Node 22.5

We have tested:

* the Docker image builds successfully
* `bin/setup` and `bin/dev` work on both Linux/x86-64 and darwin/Arm
* RSpec runs properly


spec/support/factory_bot.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.configure do |config|
+  config.include FactoryBot::Syntax::Methods
```
peteryates, line 2: I think we should call `FactoryBot.build` and `FactoryBot.create` instead of "build` and `create` - it makes it more obvious what's going on. To a newer dev I think it's a bit cryptic.

peteryates (approved):
Looks great, nice and light.

spec/support/factory_bot.rb
tonyheadford: Yes happy with that

PR #42, opened 2024-07-23
Spike email content in code

tonyheadford:
Spike using view based emails instead of template based email with Notify.
A basic template is needed in Notify with placeholders for the subject and body, then we can use maielrs and views in the code to supply the content for the email, keeping the mailer and template synchronised within the repo.

## Example Notify template
image

## Test message rendered on Notify
image

## Preview of the message using `ActionMailer::Preview` from the app
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,9 +11,12 @@ gem "pg", "~> 1.1"
 gem "propshaft"
 gem "puma", ">= 5.0"
 gem "tzinfo-data", platforms: %i[windows jruby]
+gem "dotenv-rails"
 
 gem "govuk-components"
 gem "govuk_design_system_formbuilder"
+
+gem "mail-notify"
 
 group :development, :test do
   gem "debug", platforms: %i[mri windows]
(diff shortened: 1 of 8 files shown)


claire-hughez:
This is awesome!

(Remind me to ask your favourite cheese at some point.)

Gemfile
```diff
@@ -11,10 +11,13 @@ gem "pg", "~> 1.1"
 gem "propshaft"
 gem "puma", ">= 5.0"
 gem "tzinfo-data", platforms: %i[windows jruby]
+gem "dotenv-rails"
```
peteryates, line 14: Just thinking about this, I've used [direnv](https://direnv.net/) for years because it has a few security benefits over dotenv. It's available on all platforms and works nicely with `.env` files.

For production we'll autopopulate the env from the Azure KeyVault - perhaps this isn't needed 🤔

tonyheadford: Happy to switch if there's something that is better, I've only used `dotenv` previously and it was also in ECF1. Maybe it just should be in the `:development` and possibly `:test` bundle group?

peteryates: Yeah I'm happy with either really. I do like `direnv` because it's something devs run themselves without needing a dependency.

PR #49, opened 2024-07-25
Remove placeholder title text from the general ticket template

peteryates:
When creating a new issue the 'Submit new issue' button won't be enabled until a title is entered. When there's placeholder text in the button is enabled from the start and there's a greater chance of someone (me!) submitting it with the default text.

Having a blank title forces the reporter to add something.


claire-hughez (approved):
Alright, you win.  :D

PR #52, opened 2024-07-31
Add `.editorconfig` file

peteryates:
This is largely lifted from [The GDS Way](https://gds-way.digital.cabinet-office.gov.uk/manuals/programming-languages.html) but I've condensed it to remove some repetition.

The two outliers are Makefile which needs tabs instead of spaces and SQL which is more readable with extra indentation


ltello (approved):
I like this idea too

peteryates:
Thanks @ltello.

I'll add a setup guide too, people might need an editor plugin to take advantage of this.

PR #69, opened 2024-08-05
Add quick start guide for developers and link from the README

peteryates:
This is written by someone far more familiar with Linux than macOS, suggestions/fixes would be most appreciated!

It should be in depth enough to allow a junior developer to get everything running locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,4 @@ ECF is a framework of standards to help early career teachers succeed at the sta
 ## Documentation
 
 * [Glossary](./documentation/glossary.md)
+* [Setup guide](./documentation/setup.md)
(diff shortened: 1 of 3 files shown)


steventux (approved):

steventux:
I wondered if there was a way to document necessary env vars, but this usually gets out of date. We could add an `.envrc` template which contains the required var names?

peteryates:
Ah sorry @steventux I hit merge just as your comment appeared! Yeah that's a good. I think it might be worth putting communal ones that we need to keep secret in a `dev` keyvault and add some guidance on using @saliceti's amazing [yaq](https://github.com/uk-devops/yaq) tool to download them when setting up. I'll do this as a followup.

PR #74, opened 2024-08-07
Move previews to /spec and fix bug in example

tonyheadford:
Move mailer previews to `/spec` instead of `/test` to tidy up the root folder and fix a bug in the example preview


peteryates:
I noticed this too and had a item in my todo list!

image

Don't we need to [set the default path](https://guides.rubyonrails.org/action_mailer_basics.html#previewing-emails) too though?

tonyheadford:
> 
> Don't we need to [set the default path](https://guides.rubyonrails.org/action_mailer_basics.html#previewing-emails) too though?

No it looks like the rspec install adds it for us - or at least it did on my local system

peteryates (approved):

PR #80, opened 2024-08-09
Experiment testing with Playwright

ltello:
# Playwright is a new open-source NodeJS-based browser testing framework from Microsoft. 

Some of the key features that could have an impact in our service as a replacement for Selenium and Cypress:

- Reviews say specs run faster than Selenium.
- Reviews say specs are more stable than using Selenium.
- automatic waiting for dom elements to get rendered and ready.
- works well with most CI/CD pipelines including Azure.
- supports running tests concurrently across multiple browsers (parallel testing)
- support for multiple browser engines (Chromium, WebKit, Firefox) 
- great GUI tool to trace/debug running specs
- video recording of specs run sessions for later review.
- easy to set up and run.
- very rich DSL (locators, assertions, ...)
- support multiple programming languages. For ruby we can use [playwright-ruby-client](https://playwright-ruby-client.vercel.app/) and [capybara-playwright-driver](https://rubygems.org/gems/capybara-playwright-driver)

Playwright can be set up with RSpec in several different ways:

## Setup as a capybara driver replacement for Selenium
- with this setup we basically keep using Capybara but replacing the commonly used Selenium driver with a Playwright driver provided by the gem `capybara-playwright-driver`.
- some reviews (and even the documentation of the driver) say specs are sometimes flaky and inaccurate.
- not all powerful features (auto-waiting, advanced selectors...) of Playwright can be used with this setup.

## Setup as a standalone testing tool without Capybara
- with this setup we need to manually start a Playwright browser instance and a Rails server running our app the first time a feature spec is executed on a test run.
- setting up RSpec to raise both instances above is a bit tricky but we get the benefit of being able to use all the goodies provided by Playwright including its rich DSL.
- tests run fast and accurately.

##  Setup using Capybara to raise a Rails server running our app.
- with this setup we define and register a null driver for Capybara.
- the configuration is similar to the previous approach but we take advantage of Capybara raising the Rails server with our app automatically behind the scenes the first time it is needed.

# What we need
Our service will contain very few pages with JavaScript code. 
For those pages with no JS, it is fine to test them with a browser with JS disabled so specs can run fast.

However, for pages with JS code, it is a good practice to write specs to test them in two scenarios: in a JS-enabled 
 browser and a JS-disabled one. We need to make sure these pages work well in both situations.

For this reason, we need a testing setup in our app that allow us switching the context dynamic and quickly when running the test suite. 

## What's in this PR
~This PR configs our service to use Playwright within Capybara as a null driver (method 2 above).~
This PR configs our service to use Playwright without Capybara (method 3 above).
A Chromium headless browser and a rails server serving the app is started when the first feature spec is going to be run and closed at the end of the execution of the whole suite.

## Try it
- Fetch the branch 
- locally run `bin/setup` to get the gem and node packages installed.
- Init and run the server with `bin/dev`
- in your browser navigate to `localhost:3000/countries`
- the page displayed renders an input box to select a country using autocomplete via js code 
-  type just a few chars to get a list of potential country names
- select a country

- some feature specs have been written to test this text box multiple times to test duration.
- run them like normal rspec tests: `rspec spec/features`
 

## To do
- write more specs using different locators, assertions and rest of goodies of Playwright DSL.
- investigate other potential setups 
- setup CI to install Playwright gem and JS package.


tonyheadford (commented):
Thanks for this @ltello, I agree with what @peteryates said regarding testing the pages. If we're not going to do "feature" tests then for the few things we might want to test (and we're only testing them because JS) maybe it's better that we look at the standalone option, although doing that doesn't feel a million miles away from having Cypress in ECFv1 and we didn't like the context switch 🤔 .

peteryates:
I think being able to run (and debug) everything via RSpec reduces the context switch a bit. Maybe we take a look at a standalone example and see how it feels? I'll take a closer look at this today.

tonyheadford:
> I think being able to run (and debug) everything via RSpec reduces the context switch a bit. Maybe we take a look at a standalone example and see how it feels? I'll take a closer look at this today.

Yes, that'll make a huge difference if it stays all rspec 👍

peteryates:
Thanks Lorenzo.

This looks great and everything's working for me.

I think that the example test, which checks the value on the page is good but ideally we'd submit the page and make sure Rails receives what we expect. Also it will work without having to manually run the server in the background. It seems like Capybara makes that much easier.

Of the three options in the laid out in  PR, it _seems_ like the first two would offer that. The risks of flakiness are a worry but we'll only have a few isolated pages where we run these, and the pages will need non-JavaScript fallbacks too (which we can use for the e2e/integration tests).

Shall we go with the Capybara driver for now? Eager to hear your thoughts.

ltello:
Well... I have felt really comfortable working with Playwright without Capybara. 
That made available to me the full DSL offered by Playwright which looks really great.
The tests were running at top speed. 
I could even turn headless driver off so I could see a real browser starting and executing the tests which is wonderful for debugging purposes.
Basically my experience with pure Playwright has been great!
The only downside is having to manually start/stop a Rails server, but I was able to set it up so devs have nothing to do and it would be fully transparent to them. 
If there is no issues setting the CI to work with pure Playwright I think that would be my preferred option because it sounds scary to me that even the documentation of Capybara-Playwright driver says we will have flaky and innacurate tests.

However it is also true that we expect to have little e2e/system tests and maybe going with the Playwright Capybara driver could be enough.

peteryates:
That's good enough for me @ltello, let's go with that!

ltello:
@peteryates  You mean go with capybara driver no?

spec/support/testing_with_playwright.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'rspec_playwright'
+
+RSpec.configure do |config|
+  config.add_setting :rails_server_pid
+  config.add_setting :playwright_page
+
+  # Start Rails server and Playwright browser
+  config.before(type: :feature) do
```
peteryates, line 8: Can this target JavaScript features a bit more, we shouldn't be starting Rails for regular features imo. Maybe via a tag?

spec/support/rspec_playwright.rb
```diff
@@ -0,0 +1,51 @@
...
+require 'timeout'
+require 'playwright'
+
+module RSpecPlaywright
+  PLAYWRIGHT_CLI_EXECUTABLE_PATH = "./node_modules/.bin/playwright".freeze
+  RAILS_SERVER_PROTOCOL = "http".freeze
+  RAILS_SERVER_HOST = "localhost".freeze
+  RAILS_SERVER_PORT = 3194
+  RAILS_SERVER_URL = "#{RAILS_SERVER_PROTOCOL}://#{RAILS_SERVER_HOST}:#{RAILS_SERVER_PORT}".freeze
+  RAILS_SERVER_BOOT_COMMAND = "bin/rails server -e test -p #{RAILS_SERVER_PORT}".freeze
```
peteryates, line 12: I think we can keep RSpec's output a bit quieter with:

```suggestion
  RAILS_SERVER_BOOT_COMMAND = "exec bin/rails server -e test -p #{RAILS_SERVER_PORT} > tmp/rails_playwright_last_run 2>&1".freeze
```

spec/support/testing_with_playwright.rb
peteryates: Yeah - that's it. Just so it's only started when/if necessary?

ltello: Wouldn't that mean having a DSL for one sort of tests and another for the others?
I never thought of doing it that way but let me try and see where we get.

peteryates: Even if we introduced a `playwright` [tag](https://rspec.info/features/3-12/rspec-core/command-line/tag/)?

ltello: IMO the problem is not using a `playwright` tag or any other mechanism to differenciate the 2 types of feature tests.
The issue to me is having:
- For non-JS feature tests: a technology toolbox (Capybara + Selenium + Chromium browser coming with them) with its own gems, specific RSpec setup, DSL and way to write specs, etc 
- For JS feature tests: a different technology toolbox (Playwright + Chromium browser for Playwright) with its own gems, node packages, specific RSpec setup, different DSL and different way to write specs, etc

peteryates: No, I'm suggesting that we have two types of feature test, non-JavaScript ones which just use plain old Capybara and JavaScript ones which use Playwright.

I don't think we should have any Selenium at all.

But, if I'm just running some normal non-JavaScript features I don't think it's worth starting Rails/Playwright/Chrome every time.

Issue #81, opened 2024-08-09
Job queues and workers 🧑‍💼 


peteryates:
Not everything the application does needs to be done now. Some things, like sending an email or refreshing a report might take a few seconds and we don't want to slow down the app.

For this, in Rails, we use [ActiveJob](https://guides.rubyonrails.org/active_job_basics.html).

We intend to keep ECF2 as small and simple as possible, sticking to Rails' defaults wherever possible. Rails 8 will ship with a new official queue library, [Solid Queue](https://github.com/rails/solid_queue). It uses the database as a store instead of Redis, simplifying our deployment significantly.

Solid queue supports scheduled/recurring tasks out of the box.

## Tasks

* [x] add and integrate solid_queue
* [x] create and document a basic example job
* [x] make `bin/dev` aware of it and run a worker automatically

gracecbrennan:
@peteryates / @tonyheadford to review today

PR #91, opened 2024-08-13
Add queues and workers

steventux:
Issue [81](https://github.com/DFE-Digital/ecf2/issues/81)

Adds SolidQueue a database backed async job queue

image


tonyheadford (approved):
Looks fine to me. Was wondering whether it supported cron jobs and it does, looks like they need to be added under a `recurring_tasks` section under `dispatchers` in the yaml config

PR #93, opened 2024-08-13
Added info on TRN and DoB validation

mason-emily:
This is to update sections of the services rules relating to adding details for ECTs and mentors and their training choices to reflect what we need for ECF 2.0.

### DIFF
--- a/documentation/service-rules/school-admins.md
+++ b/documentation/service-rules/school-admins.md
@@ -520,11 +520,12 @@ manual checks.
 
 Name and email address is also provided when an ECT is added. Note, we may decide to pull these details from DQT rather than asking schools for them.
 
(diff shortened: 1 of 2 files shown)


documentation/service-rules/school-admins.md
```diff
@@ -485,6 +485,12 @@ their ECTs individually.
 
 ---
 
+School must enter TRN and DoB:
+- TRN must be a 7 digit number
+- ECT must be between 18 and 100 years old
```
mason-emily, line 490: Does this age range seem sensible? We need validation to prevent typos causing date problems

peteryates: I think 100 is fine, the main objective is to prevent typos causing dates that are way out.

PR #94, opened 2024-08-13
Improve documentation site reloading

peteryates:
Use [nanoc-live](https://nanoc.app/doc/sites/#compiling-a-site) instead of relying on `entr` and specifying what files to watch. This should make working with the site when changing multiple file types (i.e., files other than markdown) a bit smoother.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,8 @@ group :test do
   gem "capybara"
 end
 
-group :documentation do
+group :nanoc do
   gem "asciidoctor"
   gem "nanoc"
+  gem "nanoc-live"
 end
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,5 @@
-watch-documentation-site:
-	ls documentation/site/content/**/* | entr -s "cd documentation/site && bundle exec nanoc"
-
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -31,7 +31,8 @@ group :test do
   gem "capybara"
 end
 
-group :documentation do
+group :nanoc do
```
peteryates, line 34: Had to change the group name here, the documentation states:

>  To set up live recompilation, add the nanoc-live gem inside the nanoc group in your Gemfile

Makefile
```diff
@@ -1,8 +1,5 @@
-watch-documentation-site:
-	ls documentation/site/content/**/* | entr -s "cd documentation/site && bundle exec nanoc"
-
 serve-documentation-site:
-	cd documentation/site/output && ruby -run -e httpd . -p 8000
+	cd documentation/site && bundle exec nanoc live --handler webrick --port 8000
```
peteryates, line 2: Using Webrick rather than Puma here because [Puma doesn't always quit cleanly with `ctrl+c`](https://github.com/nanoc/nanoc/issues/1499)

PR #97, opened 2024-08-13
Make the documentation website look less official

peteryates:
The site looked too much like a proper government service so the following changes have been made:

* use the sans-serif font family instead of Transport
* replace the crown with a giraffe
* remove the coat of arms from the footer
* add a phase banner explaining it's interal documentation

image


claire-hughez:
Any chance we can quickly change the H1 to Early career framework docs too pleaaaaase? Otherwise fabulous

EmmaFrith (approved):

PR #123, opened 2024-08-16
Add email authentication

tonyheadford:
Integrating [Omniauth](https://github.com/omniauth/omniauth) ready for DfE Sign-In and enabling email based one-time password authentication for non-DfE Sign-In capable users.  In addition we can also enable Omniauth's `Developer` strategy for `persona` sign-ins for ease of use when developing, testing and reviewing.

For personas I've made any `User` with an `@example.com` email address available to be selected.  Should be simple to tweak if we want to do something different.

I've added [ROTP](https://github.com/mdp/rotp) for OTP generation/verification as `ActiveRecord::TokenFor` was more about generating signed tokens than OTP

Added an `/admin` controller and placeholder page that is protected and you are made to sign-in to access it.  It displays info about the signed in user once successfully authenticated.

With OTP, I'm trying to handle the case when the email does not match an existing user in the DB by continuing as if it does but always failing to verify any code entered. This is to prevent leaking information about valid users.


app/views/admin/index.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
```
peteryates, line 2: I might have broken this a bit in #162 as I moved the repeated row/column stuff into the layout.

app/views/otp_sessions/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
```
peteryates, line 2: Same here re layout

peteryates (commented):
This is looking really good.

I do think that there's a lot going on in the controllers, and wonder if we might benefit from moving the functionality to a set of nice easily-testable service classes in `app/services/sessions`. It might also make sense for the corresponding forms to go in `app/forms/sessions` too.

tonyheadford:
> This is looking really good.
> 
> I do think that there's a lot going on in the controllers, and wonder if we might benefit from moving the functionality to a set of nice easily-testable service classes in `app/services/sessions`. It might also make sense for the corresponding forms to go in `app/forms/sessions` too.

Yes nice idea. I did wonder about breaking it up more in the controllers, will give it a go

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,33 @@
...
+  def authenticated?
+    current_user.present?
+  end
+
+  def session_manager
+    @session_manager ||= Sessions::SessionManager.new(session)
+  end
+
+  def login_redirect_path
+    # this should return the "home" location for the signed in user first perhaps?
```
steventux, line 30: I'd remove this and other speculative comments. They tend to become outdated, the code is a better reflection of what is actually happening.

spec/validators/notify_email_validator_spec.rb
```diff
@@ -0,0 +1,52 @@
+require "rails_helper"
+
+RSpec.describe NotifyEmailValidator do
+  # Test cases from https://github.com/alphagov/notifications-utils/blob/2432e1881cf7a5005a8d69c2ddc1597add96acc3/tests/test_recipient_validation.py
+  # The following valid emails addresses are not accepted: japanese-info@例え.テスト, info@german-financial-services.vermögensberatung
+  # This is because we do not transform to punycode
+  it "correctly identifies valid email addresses" do
+    valid_email_addresses = %w[email@domain.com email@domain.COM firstname.lastname@domain.com firstname.o\'lastname@domain.com email@subdomain.domain.com firstname+lastname@domain.com 1234567890@domain.com email@domain-one.com _______@domain.com email@domain.name email@domain.superlongtld email@domain.co.jp firstname-lastname@domain.com info@german-financial-services.reallylongarbitrarytldthatiswaytoohugejustincase email@double--hyphen.com]
```
steventux, line 8: IMO this would be better split into multiple lines.

app/services/sessions/one_time_password.rb
```diff
@@ -0,0 +1,49 @@
...
+    attr_reader :user
+
+    def initialize(user:)
+      @user = user
+    end
+
+    def generate_and_send_code
+      code = generate
+
+      Rails.logger.debug("\n\e[33;1m===>>> OTP code for #{user.email} is: [\e[36m#{code}\e[33m] <<<===\e[0m\n")
```
steventux, line 12: If we are keeping this logging it might be better to have a method which applies the formatting.

db/seeds.rb
```diff
@@ -1,9 +1,5 @@
...
-# development, test). The code here should be idempotent so that it can be executed at any point in every environment.
-# The data can then be loaded with the bin/rails db:seed command (or created alongside the database with db:setup).
-#
-# Example:
-#
-#   ["Action", "Comedy", "Drama", "Horror"].each do |genre_name|
-#     MovieGenre.find_or_create_by!(name: genre_name)
-#   end
+# some sample users/personas to test personas/otp
+User.create!(name: "Velma Dinkley", email: "velma@example.com")
```
steventux, line 2: :+1: For not using sci-fi characters

config/routes.rb
```diff
@@ -11,4 +11,22 @@
...
+  get '/sign-in', to: 'sessions#new'
+  get '/sign-out', to: 'sessions#destroy'
+
+  # one time password
+  get '/otp-sign-in', to: 'otp_sessions#new'
+  post '/otp-sign-in', to: 'otp_sessions#create'
+  get '/otp-sign-in/code', to: 'otp_sessions#request_code'
+  post '/otp-sign-in/verify', to: 'otp_sessions#verify_code'
+
+  # we won't want to do this in production but might for review apps/staging so need
```
steventux, line 26: Maybe just `# TODO: Add constraint to prevent personas working in production environment`

app/models/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ApplicationRecord
+  validates :name, presence: true
+  validates :email, presence: true, uniqueness: true, notify_email: true
+
+  # we can encrypt the secret in the DB but we need to set up keys first
```
steventux, line 5: Maybe use a `TODO:` style comment, they are easy to search for.

Gemfile.lock
```diff
@@ -448,19 +531,24 @@ DEPENDENCIES
...
+  shoulda-matchers
   solid_queue
   tzinfo-data
 
 RUBY VERSION
    ruby 3.3.4p94
 
 BUNDLED WITH
-   2.5.17
+   2.5.11
```
steventux, line 554: Do we want to decrease the bundler version?

app/views/admin/index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <p class="govuk-body">
+      Placeholder page for the admin site
+    </p>
+
+    <%= govuk_summary_list(
+      rows: [
+        { key: { text: "Signed in as:" }, value: { text:  current_user.name } },
+        { key: { text: "Email" }, value: { text:  current_user.email } },
+        { key: { text: "Via" }, value: { text:  session["user_session"]["provider"] } },
+        { key: { text: "Session expires" }, value: { text:  session["user_session"]["last_active_at"] + 2.hours } },
```
steventux, line 14: We should configure the 2 hour session duration in `SessionManager` and provide a convenient way to check/expose expiry time.

eg. `SessionManager#session_expired?` and `SessionManager#session_expires_at` or similar

app/services/sessions/one_time_password.rb
```diff
@@ -0,0 +1,49 @@
...
+    def generate
+      generate_otp_secret!
+      totp.now
+    end
+
+    def verify(code:)
+      # 600 = 10 mins validity
+      # prevent re-use within window by using the last otp_verified_at
+      params = {
+        drift_behind: 600,
```
steventux, line 28: Though it's less performant, `10.minutes.to_i` is self documenting.

steventux (commented):
This looks good to me, the only thing I think needs addressing is the session expiry time repetition.
I've made some comments about comments, meta.
The rest is just formatting nitpicking.
:100:

app/services/sessions/one_time_password.rb
peteryates:
```ruby
10.minutes.in_seconds
```

Gemfile.lock
tonyheadford: Good spot

app/views/admin/index.html.erb
tonyheadford: Yes nice idea

Gemfile.lock
```diff
@@ -121,10 +127,20 @@ GEM
       railties (>= 5.0.0)
     faker (3.4.2)
       i18n (>= 1.8.11, < 2)
+    faraday (2.10.1)
```
peteryates, line 130: Interesting that we're pulling in Faraday as there'll be some TRS API stuff coming shortly!

tonyheadford: Think that was the omniauth/open_id gems that brought that in

app/views/sessions/new.html.erb
peteryates, line 1: Doesn't really matter too much for here but [GOV.UK Components supports lists now](https://govuk-components.netlify.app/helpers/list/#plain-lists).

app/views/otp_sessions/new.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= form_for @otp_form, url: otp_sign_in_path, method: :post do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <h1 class="govuk-heading-l">
+    Sign in
+  </h1>
+
+  <p class="govuk-body">
+  You need to request an email with a code to sign in.
```
peteryates, line 9: Indentation's off a little here.

lib/colourize.rb
peteryates, line 1: Neat!

lib/colourize.rb
```diff
@@ -0,0 +1,29 @@
...
+    blue: 34,
+    magenta: 35,
+    cyan: 36,
+    white: 37,
+  }.freeze
+
+  def self.text(text, colour, mode = :bold)
+    opts = [
+      COLOURS[colour],
+      MODES[mode],
```
peteryates, line 24: Worth using `#fetch` here so it blows up?

peteryates (approved):
Looks great. Left some really minor suggestions.

app/views/sessions/new.html.erb
tonyheadford: Nice!

peteryates (approved):

PR #129, opened 2024-08-16
Use-tech-doc-template-for-docs-site

EmmaFrith:
Very much a work in progress but opening a PR early for visibility. 

Adding a navbar and beginning to break up the service rules content into smaller more readable chunks:

Screenshot 2024-08-16 at 15 33 46

This will more closely align with the [tech doc pattern](https://tech-docs.teacherservices.cloud/#teacher-services-technical-documentation) which I think is more suited to this technical and very long content.

### DIFF
--- a/documentation/site/layouts/service-rules.html
+++ b/documentation/site/layouts/service-rules.html
@@ -9,18 +9,31 @@
       <main class="govuk-main-wrapper" id="main-content">
         <div class="govuk-grid-row">
           <div class="govuk-grid-column-two-thirds">
-            <span class="govuk-caption-xl">Service rules</span>
+            <span class="govuk-caption-xl">Appropriate bodies</span>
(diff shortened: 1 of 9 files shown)


documentation/site/layouts/service-rules.html
```diff
@@ -9,19 +9,32 @@
       <main class="govuk-main-wrapper" id="main-content">
         <div class="govuk-grid-row">
           <div class="govuk-grid-column-two-thirds">
-            <span class="govuk-caption-xl">Service rules</span>
+            <span class="govuk-caption-xl">Appropriate bodies</span>
```
EmmaFrith, line 12: I want to do some logic here so that the caption shows which section in the navbar you're on

claire-hughez:
I think we decided not to do this, so should we archive and close?

PR #79, opened 2024-08-08
WIP Add AKS deployment environments with Terraform

steventux:

PR #143, opened 2024-08-21
Add staging environment

steventux:
Issue: #7 

Depends on #79 and #142 

Adds terraform, global config and make commands to apply staging environment for test cluster.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,7 +44,7 @@ jobs:
           docker-repository: ghcr.io/dfe-digital/ecf2
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
-  deploy_review:
+  deploy-review:
     name: Deploy review
     concurrency: deploy_review_${{ github.event.pull_request.number }}
     if: github.actor != 'dependabot[bot]' && github.event_name == 'pull_request'
@@ -70,3 +70,29 @@ jobs:
         with:
           message: |
             Review app deployed to ${{ steps.deploy.outputs.url }}
+
+  deploy-nonprod:
+    name: Deploy to ${{ matrix.environment }} environment
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+    concurrency: deploy_${{ matrix.environment }}
+    needs: [docker, all-checks-passed]
+    strategy:
+      max-parallel: 1
+      matrix:
+        environment: [staging]
+    environment:
+      name: ${{ matrix.environment }}
+      url: ${{ steps.deploy.outputs.environment_url }}
+    outputs:
+      environment_url: ${{ steps.deploy.outputs.environment_url }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ./.github/actions/deploy-environment-to-aks
+        id: deploy
+        with:
+          environment: ${{ matrix.environment }}
+          docker-image: ${{ needs.docker.outputs.docker-image }}
+          azure-credentials: ${{ secrets.AZURE_CREDENTIALS }}
+          current-commit-sha: ${{ github.sha }}
--- a/config/global_config/staging.sh
+++ b/config/global_config/staging.sh
@@ -1,0 +1,7 @@
+CONFIG=staging
+ENVIRONMENT=staging
+CONFIG_SHORT=st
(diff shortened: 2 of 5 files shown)


config/global_config/staging.sh
```diff
@@ -0,0 +1,7 @@
+CONFIG=staging
+ENVIRONMENT=staging
```
peteryates, line 2: It would be nice if we could make the Rails environments line up with the deployed ones, even if they do inherit from `production`.

We can base environments on other environments with:

```ruby
require Rails.root.join("config/environments/production")
```

Takes away some potential for confusion and gives a bit of extra flexibility (i.e., tweaking log levels).

peteryates: We do this in ECF1 and NPQ by assigning `RAILS_ENV` within the application config, [like this](https://github.com/DFE-Digital/npq-registration/blob/main/terraform/application/application.tf#L23).

It's nice and clean because the Rails env matches the actual environment 1:1, so if we want to adjust something somewhere (i.e., the phase banner to indicate which environment you're looking at) we can.

config/terraform/application/config/staging_Terrafile
```diff
@@ -0,0 +1,3 @@
+aks:
+    source:  "https://github.com/DFE-Digital/terraform-modules"
+    version: "main"
```
saliceti, line 3: Use "testing" in staging

config/global_config/staging.sh
```diff
@@ -0,0 +1,7 @@
...
+CONFIG_SHORT=st
+AZURE_SUBSCRIPTION=s189-teacher-services-cloud-test
+AZURE_RESOURCE_PREFIX=s189t01
+KV_PURGE_PROTECTION=false
+NAMESPACE=cpd-development
```
saliceti, line 7: Should it be cpd-test ?

steventux: I based this on NPQ Reg https://github.com/DFE-Digital/npq-registration/blob/main/global_config/staging.sh#L8 which uses `cpd-development`

.github/workflows/deploy.yml
```diff
@@ -70,3 +70,23 @@ jobs:
         with:
           message: |
             Review app deployed to ${{ steps.deploy.outputs.url }}
+
+  deploy-staging:
```
saliceti, line 74: Can we use a matrix for this environment and next ones?
As in https://github.com/DFE-Digital/itt-mentor-services/blob/main/.github/workflows/build_and_deploy.yml#L67

steventux: https://github.com/DFE-Digital/ecf2/pull/143/commits/d79bda4a0d3c6fa65e6a797712736e868e8891d5

config/terraform/application/config/staging_Terrafile
steventux: https://github.com/DFE-Digital/ecf2/pull/143/commits/c39baccb939ac0601c067df1d089d694b0976030#diff-b1187a1de37b3bbedd93597cd04224b8e016118e4a049f452964c671142cd946R3

peteryates (approved):
This looks great to me, very tidy 👌🏽

config/global_config/staging.sh
peteryates: We've used `cpd-development` in most other places (other than TSHF). Is there a problem with us using it here too @saliceti? Does it affect anything other than which `-n` we need to pass?

.github/workflows/deploy.yml
```diff
@@ -70,3 +70,29 @@ jobs:
...
+  deploy-nonprod:
+    name: Deploy to ${{ matrix.environment }} environment
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main' && github.event_name == 'push'
+    concurrency: deploy_${{ matrix.environment }}
+    needs: [docker, all-checks-passed]
+    strategy:
+      max-parallel: 1
+      matrix:
+        environment: [staging, migration. sandbox]
```
saliceti, line 83: You may want to only have [staging] for now, then add the other ones in each PR

steventux: :+1: https://github.com/DFE-Digital/ecf2/pull/143/commits/579cdf75fa306997ed5a8e6927b2273b2d533f36#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R83

PR #161, opened 2024-08-24
Minimal Playwright with Capybara

peteryates:
This is a slight twist on the work in #80 and opts to use `capybara-playwright-driver`.

The intention was to see how simple the config/setup can be for us to use Playwright. It comes with a really basic test that takes the extra step of checking that something happened (a file written); in the real app this would be checking the database for an update.


peteryates:
Closing in favour of a Playwright only approach, see #80.

PR #162, opened 2024-08-25
Layout improvements

peteryates:
This application is going to have several separate 'interfaces' for different sets of users, and different environments for internal testing, testing with providers and production.

Because of this complex set of needs we want a flexible, informative layout. Moving the bits of the main layout to partials will make it easier to create additional layouts without introducing too much duplication. Additionally, duplication is reduced between pages by moving repeating markup to the layouts.

## Custom colours per environment

To make it easier to differentiate between otherwise-identical deployments, changing the colours of the bar beneath the header and the phase banner tag makes sense. Both environment-specific changes are handled by the new EnvironmentHelper. When running in production mode the phase banner will remain blue, but the content and colours are easily customisable in the helper.

## Standard page setup (frontmatter-like)

There's now a `#page_data` helper which allows the title and header to be set in a consistent manner and simplifies the insertion of 'Error:' to the start of page titles on pages where there's an error being displayed.

Doing this in a standard way makes it much easier to maintain consistency once we have more pages by moving more functionality from every template to the layout.

## Screenshots

| Coloured header and phase banner | Optional side bar content |Setting page data |
| ------------ | ---------------- | ------- |
| image | Screenshot from 2024-08-27 00-12-36 | image |

## Review suggestions

I'd suggest reviewing commit by commit with whitespace disabled for 32e50271f08cd53fc1b49e52913729db7126b4ee.


app/views/layouts/application.html.erb
```diff
@@ -17,7 +17,19 @@
...
+          <div class="govuk-grid-column-two-thirds">
+            <%= yield(:page_header) %>
+
+            <%= yield %>
+          </div>
+          <% if content_for?(:second_column) %>
+            <div class="govuk-grid-column-one-third">
+              <%= yield(:second_column) %>
+            </div>
+          <% end %>
```
peteryates, line 31: Is there a better name for this than `:second_column`? `side_bar` maybe? 🤔

tonyheadford: sidebar sounds good 👍

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,16 @@
 module ApplicationHelper
+  def page_data(title:, header: nil, header_size: "l", error: false)
+    page_title = if error
+                   "Error: #{title}"
+                 else
+                   title
+                 end
+
+    page_header = tag.h1(header || title, class: "govuk-heading-#{header_size}")
```
tonyheadford, line 9: Will this play nice when we have forms and use the label as the heading?  (or do we have a better pattern for those?)

peteryates:
I've made some changes so now as per @saliceti's suggestion we don't use `Rails.env` at all.

@steventux, [there's now a YAML file per envonment](config/terraform/application/config/review.yml) which we can add env vars to to configure how the header looks.

ltello (approved):

PR #173, opened 2024-08-29
Make `RAILS_ENV` the same as the infrastructure env

steventux:
Part of #7 

For context https://github.com/DFE-Digital/ecf2/pull/143#discussion_r1733385305

We would like ECF v2 to behave like NPQ Registration in that `RAILS_ENV` is the same as the infrastructure env.

For review apps we need to concatenate the pull request number with the infrastructure env name to produce a unique app name.
To do this we pass `pull_request_number` to Terraform. 
For non-review environments we only need to set `ENVIRONMENT`.


steventux:
Closing because https://github.com/DFE-Digital/ecf2/pull/143#discussion_r1734753536

PR #180, opened 2024-08-29
Testing setup 1: Capybara with rack-test for JS-disabled specs and standalone Playwright for JS-enabled ones.

ltello:
# `Playwright` is a new open-source NodeJS-based browser testing framework from `Microsoft`. 

Some of the key features that could have an impact in our service as a replacement for Selenium and `Cypress`:

- Reviews say specs run faster than `Selenium`.
- Reviews say specs are more stable than using `Selenium`.
- automatic waiting for dom elements to get rendered and ready.
- works well with most CI/CD pipelines including `Azure`.
- supports running tests concurrently across multiple browsers (parallel testing)
- support for multiple browser engines (`Chromium`, `WebKit`, `Firefox`) 
- great GUI tool to trace/debug running specs
- video recording of specs run sessions for later review.
- easy to set up and run.
- very rich DSL (locators, assertions, ...)
- support multiple programming languages. For `Ruby` we can use [playwright-ruby-client](https://playwright-ruby-client.vercel.app/) and [capybara-playwright-driver](https://rubygems.org/gems/capybara-playwright-driver)

`Playwright` can be set up with `RSpec` in several different ways:

## Setup as a `Capybara` driver replacing `Selenium`
- with this setup we basically keep using `Capybara` but replacing the commonly used `Selenium` driver with a `Playwright` driver provided by the gem `capybara-playwright-driver`.
- some reviews (and even the documentation of the driver) say specs are sometimes flaky and inaccurate.
- not all powerful features (auto-waiting, advanced selectors...) of `Playwright` can be used with this setup.

## Setup as a standalone testing tool without `Capybara`
- with this setup we need to manually start a `Playwright` browser instance and a `Rails` server running our app the first time a feature spec is executed on a test run.
- setting up `RSpec` to raise both instances above is a bit tricky but we get the benefit of being able to use all the goodies provided by `Playwright` including its rich `DSL`.
- tests run fast and accurately.

##  Setup using `Capybara` to manage a `Rails` server running our app.
- with this setup we define and register a null driver for `Capybara`.
- the configuration is similar to the previous approach but we take advantage of `Capybara` managing the `Rails` server with our app automatically behind the scenes the first time it is needed.

# What we need
Our service will contain very few pages with `JavaScript` code. 
For those pages with no `JS`, it is fine to test them with a browser with `JS` disabled so specs can run fast.

However, for pages with `JS` code, it is a good practice to write specs to test them in two scenarios: in a `JS`-enabled 
 browser and a `JS`-disabled one. We need to make sure these pages work well in both situations.

For this reason, we need a testing setup in our app that allow us switching the context dynamic and quickly when running the test suite. 

# What's in this PR
This PR configs our service to use:
  - `Capybara` with `rack_test` driver to execute feature specs of pages that may or may not contain `JS` code and need to be tested in  a browser with `JS` disabled. (`rack_test` is a driver coming with `Capybara` that can't execute `JS` code)
  - `Capybara` with a null driver to execute feature specs of pages that contain `JS` code and we want to test their behaviour in a browser with `JS` enabled. Instead of via the `Capybara` driver, the browser interaction will be controlled by `Playwright` itself and will send requests to a `Rails` server running our app fully managed by `Capybara` automatically.

One important caveat of this setup is the need to use a different `DSL` for writing specs to be run with `JS` disabled (`Capybara DSL`) and those running  on a browser with `JS` enabled (`Playwright DSL`).
Due to the characteristics of our project, we would be mostly writing feature specs in Capybara DSL and a very few of them in Playwright DSL.

# Try it
- Fetch the branch 
- locally run `bin/setup` to get the gem and node packages installed.
- Init and run the server with `bin/dev`
- in your browser navigate to `localhost:3000/countries`
- the page displayed renders an input box to select a country using autocomplete capability via `JS` code 
-  type just a few chars to get a list of potential country names
- select a country
- `localhost:3000/cities` gives you a similar functionality with no `Javascript` code involved.

- some feature specs have been written to test these forms with and without `JS` enabled.
- run them like normal rspec tests: `rspec spec/features`


bin/setup
```diff
@@ -20,6 +20,11 @@ FileUtils.chdir APP_ROOT do
   # Install JavaScript dependencies
   system("npm install --nofund")
 
+  # Install Playwright
+  system(%(export PLAYWRIGHT_CLI_VERSION=$(bundle exec ruby -e 'require "playwright"; puts Playwright::COMPATIBLE_PLAYWRIGHT_VERSION.strip') &&
+           npm install playwright@$PLAYWRIGHT_CLI_VERSION || npm install playwright@next &&
+           ./node_modules/.bin/playwright install))
```
peteryates, line 26: This was a bit flaky for me locally when testing, it complained about missing libraries:

```
╔══════════════════════════════════════════════════════╗
║ Host system is missing dependencies to run browsers. ║
║ Missing libraries:                                   ║
║     libicudata.so.66                                 ║
║     libicui18n.so.66                                 ║
║     libicuuc.so.66                                   ║
║     libjpeg.so.8                                     ║
║     libwebp.so.6                                     ║
║     libpcre.so.3                                     ║
║     libffi.so.7                                      ║
╚══════════════════════════════════════════════════════╝
```

It worked for me when changing the above to:

```suggestion
  system(%(npx playwright install chromium)
```

tonyheadford: This also fixed the issue I was having on Mac

ltello: The thing is that the version of `Playwright` we install needs to be compatible with the version `playwright-ruby-client` gem installed. That's why we need to find it first.

If using `npx` is fine I can change it but the documentation said that `npx` installation was slow and recommended using `npm install` instead.

ltello: Also, the PR install several browsers. Not just Chromium. 
If we are ok with testing always with Chromium I can change it accordingly.

ltello: I wonder if this would work for you with no missing libraries:

```
  system(%(export PLAYWRIGHT_CLI_VERSION=$(bundle exec ruby -e 'require "playwright"; puts Playwright::COMPATIBLE_PLAYWRIGHT_VERSION.strip') &&
           npm install playwright@$PLAYWRIGHT_CLI_VERSION || npm install playwright@next &&
           npx playwright install chromium))
```

PR #188, opened 2024-08-30
Initial data model and GIAS import

peteryates:
This PR introduces the draft data model from the [ECF2 Rails Prototype](https://github.com/DFE-Digital/ecf2-rails-prototype) and begins the GIAS data import process.

image

## Decisions

* for now, we think moving the eligibility statuses to attributes on `School` and recalculating them whenever the GIAS school updates

## Remaining tasks

* [x] Add specs for the new models covering their relationships
* [x] Add validation to the models and test it - ensure the unique indexes are covered by Rails equivalents too
* [x] Add some enums to the models for the introduced enumerated database types
* [x] ~~Add services to calculate eligibility based on the rules~~ @mason-emily created in #125 - do this separately, it's probably going to be tricky

Refs #2, #9

### DIFF
--- a/app/models/delivery_partner.rb
+++ b/app/models/delivery_partner.rb
@@ -1,3 +1,9 @@
 class DeliveryPartner < ApplicationRecord
-  has_many :object
+  # Associations
+  has_many :provider_partnerships, inverse_of: :delivery_partner
+
+  # Validations
+  validates :name,
+            presence: true,
+            uniqueness: true
 end
--- a/app/models/delivery_partner.rb
+++ b/app/models/delivery_partner.rb
@@ -1,0 +1,3 @@
+class DeliveryPartner < ApplicationRecord
+  has_many :object
+end
--- a/app/models/ect_at_school_period.rb
+++ b/app/models/ect_at_school_period.rb
@@ -1,6 +1,48 @@
 class ECTAtSchoolPeriod < ApplicationRecord
-  has_many :induction_periods
-  has_many :mentorship_periods
-  belongs_to :school
-  belongs_to :teacher
+  include Interval
+
+  # Associations
+  belongs_to :school, inverse_of: :ect_at_school_periods
+  belongs_to :teacher, inverse_of: :ect_at_school_periods
+  has_many :induction_periods, inverse_of: :ect_at_school_period
+  has_many :mentorship_periods, inverse_of: :mentee
+  has_many :training_periods, as: :trainee
+
+  # Validations
+  validates :finished_on,
+            uniqueness: { scope: :teacher_id,
+                          allow_nil: true }
+
+  validates :started_on,
+            presence: true,
+            uniqueness: { scope: :teacher_id }
+
+  validates :school_id,
+            presence: true
+
+  validates :teacher_id,
+            presence: true
+
+  validate :school_presence
+  validate :teacher_distinct_period
+  validate :teacher_presence
+
+  # Scopes
+  scope :for_teacher, ->(teacher_id) { where(teacher_id:) }
+  scope :siblings_of, ->(instance) { for_teacher(instance.teacher_id).where.not(id: instance.id) }
+
+private
+
+  def school_presence
+    errors.add(:school_id, "School does not exist") if school.nil?
+  end
+
+  def teacher_distinct_period
+    overlapping_siblings = self.class.siblings_of(self).overlapping(started_on, finished_on).exists?
+    errors.add(:base, "Teacher ECT periods cannot overlap") if overlapping_siblings
+  end
+
+  def teacher_presence
+    errors.add(:teacher_id, "Teacher does not exist") if teacher.nil?
+  end
 end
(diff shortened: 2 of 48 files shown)


app/services/gias/schools/import.rb
```diff
@@ -0,0 +1,49 @@
+class GIAS::Schools::Import
+  def initialize(urn:, name:, local_authority:, phase:, establishment_type:, school_status:, administrative_district:, address_line1:, address_line2:, address_line3:, postcode:, primary_contact_email:, secondary_contact_email:, section_41_approved:, opened_on:, closed_on:)
```
peteryates, line 2: This can be made a bit more succinct 🙈

app/models/delivery_partner.rb
```diff
@@ -0,0 +1,3 @@
+class DeliveryPartner < ApplicationRecord
+  has_many :object
```
peteryates, line 2: ✂️

app/models/concerns/periodable.rb
```diff
@@ -0,0 +1,25 @@
+module Periodable
```
peteryates, line 1: Not sure about 'Periodable'. Maybe `DateSpan`?

app/models/ect_at_school_period.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  private
+
+  def school_siblings
+    siblings.where(school_id:)
+  end
+
+  def school_presence
+    errors.add(:school_id, "School does not exist") if school.nil?
+  end
```
peteryates, line 41: Could this be done in the `school_id`  validation above with:

```ruby
 validates :school_id,
            presence: { message: "School is missing" }
```

I don't think anyone will ever be selecting the school here so the message won't be seen by users.

app/models/ect_at_school_period.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def siblings
+    self.class.where(teacher_id:).where.not(id:)
+  end
```
peteryates, line 45: I can't think of an instance where we might use this, but really I think this kind of thing should live in the service layer as it's more business logic than database.

peteryates: Also `colleagues` is probably a more accurate name here.

db/migrate/20240830140640_import_tables.rb
```diff
@@ -0,0 +1,149 @@
+class ImportTables < ActiveRecord::Migration[7.2]
+  def change
+    create_table :academic_years do |t|
+      t.integer :year, null: false, index: { unique: true }
+
+      t.timestamps
+    end
```
tonyheadford, line 7: Can we make the primary key the year? Would make things a bit nicer I think

app/models/concerns/periodable.rb
ltello: haha I was waiting for someone English native to suggest a better name.  
I was just guessing and playful. :-)
Let's go for `DateSpan` then.

app/models/ect_at_school_period.rb
ltello: This is the validation we talked about after [this issue](https://github.com/DFE-Digital/ecf2/pull/44#pullrequestreview-2198799752).
It checks the existence of the School record associated rather than the presence of `school_id`

app/models/ect_at_school_period.rb
ltello: I was not fully happy with it either but I need it to validate new periods don't overlap with sibling ones.
I'll try and refactor it.

app/models/ect_at_school_period.rb
peteryates: Ah - I don't think we need this in addition to the presence check then. If we allowed at any point people to type in a school's ID it could happen they enter something that doesn't exist, but I think the presence check above is enough.

db/migrate/20240830140640_import_tables.rb
peteryates: I don't think we'd want auto-incrementing here, maybe something like:

```create_table(:academic_years, id: false) do |t|
  t.integer :year, options: 'primary key'
end
```

gracecbrennan:
Lorenzo is going to add a couple more and then merge for review

app/models/concerns/periodable.rb
ltello: Finally renamed to `Interval`

app/models/ect_at_school_period.rb
peteryates: @tonyheadford and I had a bit of a think about this yesterday afternoon and came up with something that might work nicely. That can be added subsequently though, it doesn't need to block. Let's have a chat about it (using date ranges).

app/models/ect_at_school_period.rb
peteryates: Yeah, because in what situation would we have a `school_id` that doesn't exist? The admin will be logged in and linked to a school, so that data will be populated without any user interaction.

ltello: Ok, I have added this validation for every association for now. Let's discuss in a chat which ones go and which ones stay.

peteryates (commented):
Looks great, excellent work.

I can't review as I opened the PR but it gets a thumbs up from me 😅

claire-hughez:
This is done, and a new PR is coming with the rest of the models.

tonyheadford (approved):
Looks great!

PR #198, opened 2024-09-03
Add terraform linting step to lint workflow

peteryates:
This change was suggested by @saliceti and is based on [the implementation from Apply For QTS](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/blob/main/.github/workflows/lint.yml#L34)

It uses `terraform fmt` and `terraform validate` to check the files are formatted properly.

Unlike Apply for QTS, who have a deployed `development` environment, we're using `staging`. As a result this relies on #143.

I've run these steps manually and they work. `terraform fmt` shows some suggested fixes but it's probably not worth fixing them yet.

~~Also @steventux, is it worth pointing this PR at #143 and rebasing the others to save applying the adjustments afterwards?~~

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -101,3 +101,29 @@ jobs:
         if: false
         run: |-
           bundle exec erblint --lint-all
+
+  terraform_linting:
+    name: Lint Terraform
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        name: Checkout Code
+
+      - name: Setup terraform
+        uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: 1.9.5
+
+      - name: Check formatting
+        run: terraform fmt -check -recursive -diff
+        working-directory: config/terraform
+
+      - name: Download vendor modules
+        run: make ci staging bin/terrafile
+
+      - name: Validate
+        run: |
+          terraform init -backend=false
+          terraform validate -no-color
+        working-directory: config/terraform/application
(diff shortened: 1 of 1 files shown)


steventux:
@peteryates do you want me to push `terraform fmt` fixes onto this branch?

.github/workflows/lint.yml
```diff
@@ -101,3 +101,29 @@ jobs:
...
+      - name: Check formatting
+        run: terraform fmt -check -recursive -diff
+        working-directory: config/terraform
+
+      - name: Download vendor modules
+        run: make ci staging bin/terrafile
+
+      - name: Validate
+        run: |
+          terraform init -backend=false
```
steventux, line 127:
```suggestion
          terraform -chdir=config/terraform/application init -backend=false
```

.github/workflows/lint.yml
```diff
@@ -101,3 +101,29 @@ jobs:
...
+          terraform validate -no-color
```
steventux, line 128:
```suggestion
          terraform -chdir=config/terraform/application validate -no-color
```

steventux (commented):
Unsure why the branch is not passing the terraform linting step. We pass the `chdir` option to terraform in our Makefile commands so perhaps this is necessary from ci workflow?
I've suggested these changes, I'm not 100% convinced they will fix the issue but they do work for me locally.

peteryates:
Seems like the `working-directory` setting was causing problems in the final step, hoping 795f05aa028f1ae5a26f40d5bd21b507307e48b1 does the trick.

steventux (approved):

PR #210, opened 2024-09-05
Only deploy review apps on PRs explicitly labelled with 'deploy'

steventux:
Amends the github deploy workflow to detect the presence of the label 'deploy' before creating a review app deployment.
Every little helps.


peteryates:
I like the idea of this 👍🏽

saliceti (approved):

PR #222, opened 2024-09-06
Add the fields needed for AB data submission

### LINKED ISSUE #204: Add the required tables, fields and appropriate model validations
The data we will receive from appropriate bodies via CSV uploads and what it relates to is:

* the appropriate body:
    * `Establishment ID`
* the ECT:
    * `TRN`
    * `Surname`
    * `First Name`
    * `Date of Birth`
    * `Induction Outcome`
* an induction period:
    * `Induction Programme Type`
    * `Induction Period Start Date`
    * `Induction Period End Date`
    * `Induction Period Number Of Terms`

We intend to record data in the following process:

1. we add incoming teacher/induction data to a staging area where it's given a `pending` status
2. a background job is triggered for each row, it checks the eligibility of the teacher against the TRS API. We probably want to check for duplicates too (i.e. another row for the teacher with the same start date). The result should be one of the following:
    * **eligible** - we persist the data to the main schema and delete the staging row
    * **ineligible** - we leave the staging row so the AB can see that something didn't go to plan and hopefully give a constructive reason why it didn't. We might want to let them tweak the data in the staging area to fix a typo or something before it's resubmitted
    * **failure** - the API couldn't be reached, there was a bug, etc. Add a message/event saying what happened and schedule a retry.

## Tasks

* [x] create a _staging_ table (come up with a better name) that holds the submitted data specified above
* [x] add the fields to the database that will hold the persisted data:
    * [x] `appropriate_bodies.local_authority_code`
    * [x] `appropriate_bodies.establishment_number` (these two combined make the `EstablishmentID`)
    * [x] `teachers.trn`
    * [x] ~~`teachers.induction_outcome` (think about this first!)~~
    * [x] `induction_periods.programme_type` (will need an enum)
    * [x] `induction_periods.number_of_terms`

We won't persist `Surname`, `Firstname` or `Date of Birth`

peteryates:
Adds the fields and tables suggested in #204  

- **Add AB portal fields to the appropriate_body**
- **Add TRN to teacher model**
- **Add induction programme and number of terms fields**
- **Add pending induction submissions**

I recommend reviewing commit-by-commit as the messages explain the approach.

Fixes #204

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -38,3 +38,6 @@ Style/SignalException:
 
 Style/StringConcatenation:
   Enabled: false
+
+Rails/UniqueValidationWithoutIndex:
+  Enabled: false
--- a/app/models/teacher.rb
+++ b/app/models/teacher.rb
@@ -1,4 +1,6 @@
 class Teacher < ApplicationRecord
+  TRN_FORMAT = %r{\A\d{7}\z}
+
   # Associations
   has_many :ect_at_school_periods, inverse_of: :teacher
   has_many :mentor_at_school_periods, inverse_of: :teacher
@@ -6,4 +8,8 @@ class Teacher < ApplicationRecord
   # Validations
   validates :name,
             presence: true
+
+  validates :trn,
+            format: { with: TRN_FORMAT, message: "TRN must be 7 numeric digits" },
+            uniqueness: { message: 'TRN already exists', case_sensitive: false }
 end
--- a/db/migrate/20240904125324_add_establishment_and_la_fields_to_abs.rb
+++ b/db/migrate/20240904125324_add_establishment_and_la_fields_to_abs.rb
@@ -1,0 +1,10 @@
+class AddEstablishmentAndLaFieldsToAbs < ActiveRecord::Migration[7.2]
+  def change
+    # rubocop:disable Rails/NotNullColumn, Rails/BulkChangeTable
+    add_column :appropriate_bodies, :local_authority_code, :integer, null: false
+    add_column :appropriate_bodies, :establishment_number, :integer, null: false
+    add_column :appropriate_bodies, :establishment_id, :virtual, type: :string, as: %(local_authority_code::varchar || '/' || establishment_number::varchar), stored: true
+    add_index :appropriate_bodies, %i[local_authority_code establishment_number], unique: true
+    # rubocop:enable Rails/NotNullColumn, Rails/BulkChangeTable
+  end
+end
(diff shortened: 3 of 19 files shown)


db/migrate/20240904125324_add_establishment_and_la_fields_to_abs.rb
```diff
@@ -0,0 +1,10 @@
+class AddEstablishmentAndLaFieldsToAbs < ActiveRecord::Migration[7.2]
+  def change
+    # rubocop:disable Rails/NotNullColumn, Rails/BulkChangeTable
+    add_column :appropriate_bodies, :local_authority_code, :integer, null: false
+    add_column :appropriate_bodies, :establishment_number, :integer, null: false
+    add_column :appropriate_bodies, :establishment_id, :virtual, type: :string, as: %(local_authority_code::varchar || '/' || establishment_number::varchar), stored: true
```
peteryates, line 6: [Rails has supported generated fields in PostgreSQL since version 7](https://blog.saeloun.com/2022/01/25/rails-7-postgres-support-for-generated-columns/).

db/schema.rb
```diff
@@ -30,6 +31,10 @@
     t.string "name", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "local_authority_code", null: false
+    t.integer "establishment_number", null: false
+    t.virtual "establishment_id", type: :string, as: "((((local_authority_code)::character varying)::text || '/'::text) || ((establishment_number)::character varying)::text)", stored: true
```
peteryates, line 36: This is failing in CI. Maybe a version mismatch?

peteryates: Yep, upgraded to version 16 in 4e18c4f53b63670d55adf71dfde806 and all's good.

.rubocop.yml
```diff
@@ -38,3 +38,6 @@ Style/SignalException:
 
 Style/StringConcatenation:
   Enabled: false
+
+Rails/UniqueValidationWithoutIndex:
+  Enabled: false
```
peteryates, line 43: This wasn't playing ball when the unique index spanned multiple columns.

app/models/teacher.rb
```diff
@@ -6,4 +6,8 @@ class Teacher < ApplicationRecord
   # Validations
   validates :name,
             presence: true
+
+  validates :trn,
+            inclusion: { in: 999_999..10_000_000, message: 'TRN must be a 7 digit number' },
```
tonyheadford, line 11: This allows 999_999 in the range
```suggestion
            inclusion: { in: 1_000_000..10_000_000, message: 'TRN must be a 7 digit number' },
```

peteryates:
TRNs which were numeric have been replaced with strings. Despite [the guidance](https://www.gov.uk/guidance/teacher-reference-number-trn) saying they're a 7 digit number, they're actually a 7 digit string. 0 padded values are accepted. 🙈

tonyheadford (approved):

PR #234, opened 2024-09-09
Use date ranges to check period overlaps

peteryates:
This change adds some generated fields to the periods tables that hold a [PostgreSQL range](https://www.postgresql.org/docs/current/rangetypes.html) that spans the `started_on` and `finished_on` boundaries.

These fields are taken care of by PostgreSQL automatically but fully supported by Rails.

Ranges give us much more power than just storing two dates. They allow us to easily:

* check for overlaps for validation purposes
* combine many records to calculate entire durations
* overlap multiple kinds of periods (i.e., to easily work out how much time someone spent with an open `training_period` and `induction_period`)

I've also replaced the existing `.overlapping` method which takes two dates with an `.overlaps_with` scope that takes another period. This approach means we can query the database for clashes without having to return any data and instantiate objects, while remaining readable. 

**This needs to follow #222 which bumps the PostgreSQL version used in CI.**

### DIFF
--- a/app/models/concerns/interval.rb
+++ b/app/models/concerns/interval.rb
@@ -8,18 +8,12 @@ module Interval
     # Scopes
     scope :extending_later_than, ->(date) { where("finished_on IS NULL OR finished_on > ?", date) }
     scope :starting_earlier_than, ->(date) { where("started_on < ?", date) }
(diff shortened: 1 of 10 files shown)


tonyheadford (approved):
Nice!

PR #251, opened 2024-09-12
Replace devcontainer with Rails-generated one

peteryates:
This is the default Rails devcontainer setup minus Redis which we're not using (yet).


peteryates:
Closed in favour of #258

PR #258, opened 2024-09-13
Clean up Codespaces config

peteryates:
Remove some unneeded stuff from the Codespaces config and based it on Ubuntu 24.04 instead of 22.04.


peteryates:
@EmmaFrith mind taking this for a spin?

EmmaFrith:
When opening a codespace I get this message:

Screenshot 2024-09-13 at 11 19 21

Presumably this indicates a problem?

peteryates:
That is a problem but it's not related to the building of the Codespace. Ignore it and we'll fix that next!

Does the app run when you do `bin/dev`?

Issue #265, opened 2024-09-15
Appropriate body claim flow


peteryates:
The claim flow allows an appropriate body to _claim_ an ECT. It's fully outlined in #253.

## Done in #272

* [x] build the controllers and pages
    * [x] **Find** an ECT against the DQT
    * [x] **Check** their details are correct
    * [x] **Claim** them for the AB
    * [x] **Confirm** they were claimed

## Done in #334 

* [x] test the happy path with a feature
* [x] request specs to make sure the right services receive the right params
* [x] unit tests for the services, validations, etc

## Relies on (out of scope)

* #241 - to make sure our logged in user can be associated with an appropriate body
* #288 - TRS API integration (just hardcode a response for now and iterate when the API client is built)

PR #272, opened 2024-09-16
Add the appropriate body flow for claiming an ECT

peteryates:
As detailed in #253 and #265, this change introduces the claiming of an ECT by an appropriate body.

The purpose isn't for the flow to be _complete_, but to build something minimal that we can iterate. As a result, it doesn't bother with any API connections and leaves a sprinkling of `# FIXME` comments here and there 🙂 

## Finding the ECT in the DQT via the TRS API

| No errors | Errors |
| -------- | ----------- |
| Screenshot from 2024-09-16 17-00-20 | Screenshot from 2024-09-16 17-00-28 |

### Checking the details returned from the DQT are what we expect

Note there's more information to add to this page, that will be addressed later.

| No errors | Errors |
| -------- | ----------- |
| Screenshot from 2024-09-16 17-01-08 | Screenshot from 2024-09-16 17-01-14 |

### Adding more details about the ECT's induction

| No errors | Errors |
| -------- | ----------- |
| Screenshot from 2024-09-16 17-08-01 | Screenshot from 2024-09-16 17-01-40 |

### Done

Screenshot from 2024-09-16 17-01-48

## Notes

The designs are still a WIP and there's additional data to be added to some pages (the check data page in particular).

I'd recommend reviewing commit-by-commit.


config/initializers/date_formats.rb
```diff
@@ -0,0 +1,4 @@
+# https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates
+Date::DATE_FORMATS[:govuk]        = "%-d %B %Y" # 2 January 1998
+Date::DATE_FORMATS[:govuk_short]  = "%-d %b %Y" # 2 Jan 1998
+Date::DATE_FORMATS[:govuk_approx] = "%B %Y"     # January 1998
```
peteryates, line 4: I'm going to raise a PR on the [rails template](/dfe-digital/rails-template) to add these by default, I suspect every service reimplements them.

tonyheadford (approved):
Looks great

kas-madrigal:
Food for thought on the claim journey and how we ask for induction start date from Nathan's feedback on my prototype. @emily-prudence-dfe @peteryates 

- ABs confused the start date that the ECT starts with them vs induction in general. We only want them to tell us when the ECT started with _them_ not induction in general, the general calculation is the accumulation of those dates, hence we don't need their overall induction start date. 
- Could we/should we ask for predicted end date? 
- Can we create a fake concept of breaks that are recorded as starts and stops? i.e. can we allow for a journey that helps them record a 'break' (and use this word) but it is recorded as a stop on our side? 

These are changes that would still be alive post AB portal decommission, so actually might be worth considering them.

peteryates:
> ABs confused the start date that the ECT starts with them vs induction in general. We only want them to tell us when the ECT started with them not induction in general, the general calculation is the accumulation of those dates, hence we don't need their overall induction start date.

When asking for that date we say (in this early un-content-designed prototype): **When did Kirk Van Houten start their induction with you?**

Is that enough? Could we reword that to make it more explicit?

> Can we create a fake concept of breaks that are recorded as starts and stops? i.e. can we allow for a journey that helps them record a 'break' (and use this word) but it is recorded as a stop on our side?

I believe we've provisioned for that already, it'll be covered by this issue #329

Here's the graphic we came up with what we think are all the scenarios and triggers (yellow boxes) for sending updates through to TRS.

image

> Could we/should we ask for predicted end date?

No reason why we can't - maybe it's a post MVP thing though?

emily-prudence-dfe:
`> ABs confused the start date that the ECT starts with them vs induction in general. We only want them to tell us when the ECT started with them not induction in general, the general calculation is the accumulation of those dates, hence we don't need their overall induction start date.`
> 
> When asking for that date we say (in this early un-content-designed prototype): When did Kirk Van Houten start their induction with you?
> 
> Is that enough? Could we reword that to make it more explicit?

This is a not for now (I'll create another ticket) but one way we could make this clearer is if we did something like this:

image

Issue #241, opened 2024-09-10
User roles


peteryates:
Several different kinds of people will use the service, including:

* School admins (formerly known as SITs)
* DfE staff (admin, finance)
* Appropriate body users

When they log in we want to ensure they are directed to the right place and can only see what they're meant to see.

## Tasks

* [ ] Add separate layouts per user type (school admins, dfe admins, appropriate bodies)
* [ ] Ensure people who sign in with a particular role are directed to the right area and can only see that area
* [ ] We need to be able to grant users roles - discuss this before implementing!
* [ ] Extend the personas so they cover the different roles
* [ ] Add a secondary navigation menu so DfE admins can drill into specific areas

steventux:
## Users & roles
 
#### What questions do we ask of the authorisation service?

- What type of user is this?
- Can this user access data for this AB or School?
- Which ABs or Schools does this user have access to?
- Which users have access to this AB or School?
- Is this an admin or super admin or finance user?
- What are all the roles this user has?
- What are all the possible user roles?

#### Additional more general questions

- How do DfE staff authenticate?
- DfE Sign In usually still requires that we persist some user data, these will typically be school admin users. Do school admins have varying levels of access?

#### Technical notes

We will probably have several user types, we may eventually have several organisation types.
Regardless of this we should have one common service-like authorisation object that we can ask the above questions.

claire-hughez:
Get something out for people to talk about that is just a draft.

claire-hughez:
This morning will try and get a draft out

claire-hughez:
Pushed up draft, thought about ways to authorise users. Up for discussion, trying to lock down admin controller as proof of concept. Steve will push up those commits soon.

gracecbrennan:
PR is open. 

Steve is going to take a look at one of the last parts for when a user logs in and we find out who they are.

PR #275, opened 2024-09-17
Add user roles

steventux:
Part of #241 

#### Adds association models for users and roles. 

We treat the association between eg. a School and a User as the required authorisation for that user to perform common school administrator tasks for the associated school.

There are 5 different user roles: 
- DfE user: either super admin, finance or admin user, the default is admin. These have the most advanced access, typically via the admin interface.
- Lead provider user: These users have an association to a lead provider, they may see additional configuration options for that provider. How their view of the service differs from other users is TBC.
- Delivery partner user: Similar to lead provider users, there will be additional configuration available.
- Appropriate body user: These users have an association with an appropriate body, they will see a different view of the service from school users.
- School user: These school administrators will manage ECTs and Mentors for their school using the appropriate service interface.

#### Authorisation mechanism

I've added a concern to `User` model with helpers to determine general user access and access to specific resources.
This is a best guess as we don't have all the relevant user journeys built yet.

Also there is a controller concern to check user access, for now this is only implemented for users with one or more `DfERole`  via `AdminController`

#### Admin UI for user role management

This doesn't currently handle assigning DfE user roles which are different from model association roles. We should probably restrict that to super admins and handle elsewhere.

GET `/admin/users/:id/roles`
image

Assigning roles by type:
image

**Metaprogramming Warning**

All the association role classes are the same shape and do the same thing, and we agreed to avoid polymorphic associations. The admin roles form and relevant views contain a bit of metaprogramming to DRY up persisting, inflating and rendering roles for a user. It may be preferable to be explicit.

### DIFF
--- a/app/forms/admin/roles_form.rb
+++ b/app/forms/admin/roles_form.rb
@@ -1,0 +1,33 @@
+module Admin
+  class RolesForm
+    include ActiveModel::Model
+
+    ROLE_TYPES = %w[appropriate_body delivery_partner lead_provider school].freeze
+
+    attr_accessor :role_type, :user, :role_ids
+
+    validates :role_type, inclusion: { in: ROLE_TYPES }
+    validates :user, presence: true
+
+    def collection_for_role_type
+      role_type_class.all
+    end
+
+    def save!
+      roleables = role_ids.compact_blank.map do |role_id|
+        role_type_class.find(role_id)
+      end
+      user.send("#{role_type.pluralize}=", roleables)
+    end
+
+    def role_ids
+      @role_ids ||= user.send(role_type.pluralize).map(&:id)
+    end
+
+  private
+
+    def role_type_class
+      role_type.classify.constantize
+    end
+  end
+end
--- a/app/forms/admin/roles_form.rb
+++ b/app/forms/admin/roles_form.rb
@@ -4,7 +4,8 @@ module Admin
 
     ROLE_TYPES = %w[appropriate_body delivery_partner lead_provider school].freeze
 
-    attr_accessor :role_type, :user, :role_ids
+    attr_accessor :role_type, :user
+    attr_writer :role_ids
 
     validates :role_type, inclusion: { in: ROLE_TYPES }
     validates :user, presence: true
--- a/app/models/concerns/authorisable.rb
+++ b/app/models/concerns/authorisable.rb
@@ -15,12 +15,19 @@ module Authorisable
 
   def can_access?(model)
     return true if dfe_user?
-    return AppropriateBodyRole.where(user: self, appropriate_body: model).exists? if model.is_a?(AppropriateBody)
-    return DeliveryPartnerRole.where(user: self, delivery_partner: model).exists? if model.is_a?(DeliveryPartner)
-    return LeadProviderRole.where(user: self, lead_provider: model).exists? if model.is_a?(LeadProvider)
-    return SchoolRole.where(user: self, school: model).exists? if model.is_a?(School)
 
-    false
+    case model
+    when AppropriateBody
+      AppropriateBodyRole.where(user: self, appropriate_body: model).exists?
+    when DeliveryPartner
+      DeliveryPartnerRole.where(user: self, delivery_partner: model).exists?
+    when LeadProvider
+      LeadProviderRole.where(user: self, lead_provider: model).exists?
+    when School
+      SchoolRole.where(user: self, school: model).exists?
+    else
+      false
+    end
   end
 
   def access_type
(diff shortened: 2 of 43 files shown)


spec/rails_helper.rb
```diff
@@ -4,6 +4,9 @@
 
 abort("The Rails environment is running in production mode!") if Rails.env.production?
 require 'rspec/rails'
+require "capybara/rspec"
+
+Capybara.server = :puma
```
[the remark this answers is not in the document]
peteryates: As discussed, let's leave this for now. Capybara is probably going to be needed because when driving an actual browser the tests and browser are in separate transactions and it makes putting the right data in the right place really difficult.

app/forms/admin/roles_form.rb
```diff
@@ -0,0 +1,34 @@
...
+  class RolesForm
+    include ActiveModel::Model
+
+    ROLE_TYPES = %w[appropriate_body delivery_partner lead_provider school].freeze
+
+    attr_accessor :role_type, :user
+    attr_writer :role_ids
+
+    validates :role_type, inclusion: { in: ROLE_TYPES }
+    validates :user, presence: true
```
peteryates, line 11: I feel safer with the validations being on the model; if we're importing users and roles from ECF1 for example we would have to bring them in via this form object.

peteryates: However, we can shift things around later, for now it's fine.

app/models/concerns/authorisable.rb
```diff
@@ -0,0 +1,44 @@
...
+    has_many :lead_providers, through: :lead_provider_roles
+    has_many :school_roles
+    has_many :schools, through: :school_roles
+    has_many :dfe_roles
+  end
+
+  def can_access?(model)
+    return true if dfe_user?
+
+    send("#{model.class.name.underscore}_roles").where("#{model.class.name.underscore}": model).exists?
```
peteryates, line 19: Interesting approach. Definitely neat from a dev point of view but harder to `grep` for!

peteryates (approved):
Looks good. Going to have a think about validations and how that side of things will work.

app/models/concerns/authorisable.rb
steventux: This one could/should just be a case statement.

app/forms/admin/roles_form.rb
peteryates: Yeah, that makes sense. I think it might make sense to leave this as is for now and once have a few more features in place maybe look as a team at how we want to organise them.

app/models/concerns/authorisable.rb
steventux: https://github.com/DFE-Digital/ecf2/pull/275/files#diff-4619408be7b9b996c7394a6c5bac2a3996099036740f8f673ec9934c21a7c26cR16-R24

app/models/concerns/authorisable.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def can_access?(model)
+    return true if dfe_user?
+    return AppropriateBodyRole.where(user: self, appropriate_body: model).exists? if model.is_a?(AppropriateBody)
+    return DeliveryPartnerRole.where(user: self, delivery_partner: model).exists? if model.is_a?(DeliveryPartner)
+    return LeadProviderRole.where(user: self, lead_provider: model).exists? if model.is_a?(LeadProvider)
+    return SchoolRole.where(user: self, school: model).exists? if model.is_a?(School)
+
+    false
+  end
```
peteryates, line 24: Much easier to follow. Would a `case` statement be even clearer?

```suggestion
  def can_access?(model)
    case model
    when AppropriateBody
       AppropriateBodyRole.where(user: self, appropriate_body: model).exists?
    when DeliveryPartner
       DeliveryPartnerRole.where(user: self, delivery_partner: model).exists?
    when LeadProvider
      LeadProviderRole.where(user: self, lead_provider: model).exists?
    when School
       SchoolRole.where(user: self, school: model).exists? 
    else
      dfe_user?
  end
```

app/views/admin/roles/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    end
+  ) %>
+<% else %>
+  <p class="govuk-body">No roles assigned</p>
+<% end %>
+
+<%= form_for @roles_form, url: role_type_admin_user_path(@user), method: :get do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <%= f.govuk_collection_radio_buttons :role_type, Admin::RolesForm::ROLE_TYPES, :to_s, :humanize, legend: { text: "Choose role type" } %>
```
peteryates, line 18: It never occurred to me to use `:to_s` and `:humanize` on an array like this! 💡

app/models/concerns/authorisable.rb
peteryates: Oops sorry my indentation was a bit skewiff in that suggestion 🙈

PR #292, opened 2024-09-19
Add Capybara for non-js features

steventux:
Following a discussion about Playwright we've decided to add Capybara as a quick way to get non-JS feature specs up and running. We may be able to use the [Capybara Playwright driver](https://github.com/YusukeIwaki/capybara-playwright-driver) in future.


peteryates (approved):

peteryates:
Nice, all working perfectly for me.

PR #302, opened 2024-09-20
Migration kickoff

tonyheadford:
Adding the basic framework and a few simple migrations to show how we can migrate the data from ECF to ECF2.

This has been **heavily** borrowed from the [NPQ Separation](https://github.com/DFE-Digital/npq-registration/tree/main/app/services/migration) work and follows their approach.

**Outstanding issues:**
 * Sort out how to test migrators - we will need a database to represent ECF1
 * Sort out a good way to document the process and migrators

**Bonus 🎉** 
Added a copy of the NPQ Separation UI for migrations to run/display results

image


app/migration/migrators/academic_year.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def self.reset!
+      if Rails.env.development?
+        ::AcademicYear.connection.execute("TRUNCATE #{::AcademicYear.table_name} RESTART IDENTITY CASCADE")
+      end
+    end
+
+    def migrate!
+      Rails.logger.info("Migrating #{Migration::Cohort.count} cohorts")
```
peteryates, line 23: Should this be calling the migration logger (`logger`/`@logger`) instead of `Rails.logger`?

app/views/migration/migrations/_in_progress_migration.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <div class="govuk-notification-banner__header">
+    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
+      Important
+    </h2>
+  </div>
+  <div class="govuk-notification-banner__content">
+    <h3 class="govuk-notification-banner__heading">
+      A migration is currently in-progress
+    </h3>
+    <p class="govuk-body">It was started <%= tag.strong(time_ago_in_words(migration_started_at(@data_migrations))) %></T=> ago.</p>
```
peteryates, line 11:
```suggestion
    <p class="govuk-body">It was started <%= tag.strong(time_ago_in_words(migration_started_at(@data_migrations))) %> ago.</p>
```

app/views/migration/migrations/index.html.erb
```diff
@@ -0,0 +1,14 @@
+<% page_data(title: "Migrations") %>
+
+<% if @starting_migration || @in_progress_migration %>
+  <%= render(partial: "in_progress_migration") %>
+<% end %>
+
+<p class="govuk-body">You can run the migration from ECF using this migration tool.</p>
+
+<%= button_to "Run migration", { action: :create }, class: "govuk-button govuk-button--warning", disabled: @data_migrations.present? %>
```
peteryates, line 9: I think [govuk_button_to](https://govuk-components.netlify.app/helpers/button/#regular-button) should work here, with `warning: true`.

app/migration/migrators/academic_year.rb
tonyheadford: Following the NPQ separation approach it's more difficult to pass this in when they are running as jobs. If we want it I think we'll have to revisit the pattern a bit.

peteryates: Ah, fair.

peteryates (approved):

Issue #50, opened 2024-07-25
Importing school data 🏫 


peteryates:
While the final data model isn't yet finalised we **know** we'll need a table containing school data from GIAS.

Essentially, this time around, we intend on having two 'school' models:

* `School` - our representation of a school within the system
* `GIAS::School` - a direct import from GIAS that uses `urn` as primary key

The separation will allow us to deal with things like [the academisation of schools](https://www.gov.uk/guidance/convert-to-an-academy-information-for-schools) by simpliy switching the URN on the `School`, rather than having to migrate all the attached stuff over to a new school record. We'll just concentrate on `GIAS::School` for now though.

## Tasks

* [x] Work out which columns we need - I suspect we can start with `urn`, `name`, `section_41_approved`
* [x] Create a minimal `GIAS::School` model with the identified fields
* [ ] Create a task that pulls GIAS data and imports it. As we're likely to be using DfE Sign-in this time around let's try starting off with this [dfe-data](https://github.com/DFE-Digital/gias-data) export first, it makes the process much easier by taking care of the CSV processing and gives a nice JSON file we can just pull in. We only need open schools so ignore the closed ones.
* [ ] Add a service class responsible for doing the pull/import
* [ ] Add a rake task that performs the import

## Bonus stretch task

* [ ] #337
    * Predecssor
    * Successor
    * Predecessor - merged
    * Successor - merged
    * Successor - amalgamated
    * Predecessor - amalgamated
    * Merged
    * Plus quite a few more... 🤷🏽 

## Future notes

At some point in the future we're likely to want the email addresses/contact info from the GIAS API. We may want to create a slightly different import to pull that info.

tonyheadford:
I've done a bunch of this, with discussions with @peteryates but still ongoing

tonyheadford:
I've built quite a bit more of this and there are a couple of things worth noting:
* We found the JSON approach with that amount of data _much_ slower than reading and processing the GIAS CSV so have gone with the CSV
* I've used a different approach to the service class based upon the principle of a class with related public service methods instead of the one class per service with a `call` method and associated magic to make it work as a class method as used in ECF v1.
* The import and basic update from GIAS is there, the handling of closures/mergers/splits etc has not been done as yet

claire-hughez:
Working on other ticket on testing, will pick this up if there's time.

claire-hughez:
Tony and Lorenzo to catch up on this. Sticking with CSV files.

[...]

PR #305, opened 2024-09-23
Add GIAS::School and GIAS::SchoolLink models

ltello:
Add GIAS::School and GIAS::SchoolLink models with validations ready for the first import of schools from GIAS service


tonyheadford (commented):
Wondering whether the concerns (helpers and types) are necessary in this form now?
In ECF1 I added them because they were shared between the `::School` and `DataStage::School` models but here they are only really relevant to our `GIAS::School` certainly the scopes are 🤔

db/schema.rb
```diff
@@ -95,26 +95,47 @@
...
     t.date "closed_on"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "administrative_district_code", null: false
+    t.string "administrative_district_name"
+    t.integer "easting", null: false
+    t.integer "local_authority_code", null: false
+    t.string "local_authority_name"
+    t.integer "northing", null: false
+    t.integer "number", null: false
```
tonyheadford, line 129: We need a better name here than `number`

db/schema.rb
```diff
@@ -95,26 +95,47 @@
...
     t.string "postcode"
     t.string "primary_contact_email"
     t.string "secondary_contact_email"
-    t.boolean "section_41_approved"
     t.date "opened_on"
     t.date "closed_on"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "administrative_district_code", null: false
+    t.string "administrative_district_name"
```
tonyheadford, line 124: We were trying to remove the code/name pairings where possible and have a single value just for one or the other (whatever made the most sense) - probably the `xx_name` is going to make the most sense when looking at the data for most of the pairings?

app/models/gias/school_link.rb
```diff
@@ -0,0 +1,41 @@
...
+  # Associations
+  belongs_to :from_gias_school, class_name: "GIAS::School", foreign_key: :urn, primary_key: :urn, inverse_of: :gias_school_links
+  belongs_to :to_gias_school, class_name: "GIAS::School", foreign_key: :link_urn, primary_key: :urn
+
+  # Validations
+  validates :link_date,
+            presence: true
+
+  validates :link_type,
+            inclusion: { in: LINK_TYPES }
```
tonyheadford, line 33: I don't know if we can guarantee this list won't change. I don't think it's an issue if we just record whatever value GIAS supplies us with?  So maybe `link_type` just needs to be a `presence: true` validation?

db/schema.rb
ltello: The original name was "establisment_number". 
I can bring that one back.

app/models/gias/school_link.rb
ltello: Fine by me but if we are adding code to process automatically these "operations" in the fufure, isn't is better if the import fails (which will warn us about a new un-treatable LINK_TYPE coming in the pull) instead of silently store it?

tonyheadford:
> > Wondering whether the concerns (helpers and types) are necessary in this form now? In ECF1 I added them because they were shared between the `::School` and `DataStage::School` models but here they are only really relevant to our `GIAS::School` - certainly the scopes are only relevant to that 🤔
> 
> I just left the concerns as they were in the prototype/ECF1 because we have no fields yet in the `School` model and until we decide which ones to add. However, thinking about it now, maybe the scope` :eligible_for_registration` will only apply to `GIAS::School` and not `School`. The rest I think will be useful for both models?
> 
> When it comes to `GIAS::TYPES` it seems to only apply to `GIAS::School` but shall we leave it in their own module to separate it from the `GIAS::School` model itself?

I don't think it will hurt to leave it at the moment so that we can get this in but given the `::School` is a shell object and all the data is on `GIAS:School` I can't imagine much of the scopes and methods will be the same for both?  I'd go with as simple as possible for everything at the moment and add what we need as we find we need it?

db/schema.rb
ltello: Yes, initially I thought about leaving only the `*_code` fields to prevent inconsistencies between the 2 fields. 

However, as we don't do anything with the `*_name` (or even have validations checking that `*_code` is congruent with the `*_name`) I thought it doesn't hurt to have everything as it is served by GIAS (in case it might be helpful in the future to detect wrong data coming from GIAS).

ltello:
> > > Wondering whether the concerns (helpers and types) are necessary in this form now? In ECF1 I added them because they were shared between the `::School` and `DataStage::School` models but here they are only really relevant to our `GIAS::School` - certainly the scopes are only relevant to that 🤔
> > 
> > 
> > I just left the concerns as they were in the prototype/ECF1 because we have no fields yet in the `School` model and until we decide which ones to add. However, thinking about it now, maybe the scope` :eligible_for_registration` will only apply to `GIAS::School` and not `School`. The rest I think will be useful for both models?
> > When it comes to `GIAS::TYPES` it seems to only apply to `GIAS::School` but shall we leave it in their own module to separate it from the `GIAS::School` model itself?
> 
> I don't think it will hurt to leave it at the moment so that we can get this in but given the `::School` is a shell object and all the data is on `GIAS:School` I can't imagine much of the scopes and methods will be the same for both? I'd go with as simple as possible for everything at the moment and add what we need as we find we need it?

Ah! ok. I was with the ECF1 mindset. 
So the idea is to not replicate the `GIAS::School` fields into `School` but delegate it (which is much nicer).

tonyheadford:
> Ah! ok. I was with the ECF1 mindset. So the idea is to not replicate the `GIAS::School` fields into `School` but delegate it (which is much nicer).
Lets wait and see what we need to do regarding that functionality, it might be that all the logic is encapsulated in service objects and we won't need to `delegate` anything

app/models/concerns/gias/types.rb
peteryates, line 1: I think this can live directly in `GIAS::School`.

db/schema.rb
peteryates: I think having the code **and** name is a bit much. I think it makes sense just to bring the names in.

ltello: We can't get rid of the `administrative_district_code` because it is used to detect if the school is English/Welsh, etc.

app/models/concerns/gias/types.rb
ltello: Like `GIAS::School`, `GIAS::School:Importer` class (in the next PR) heavily depends on `GIAS::School::Types`

db/schema.rb
peteryates: Yeah. The minimum of what we know we need seems like a good starting point.

app/models/concerns/gias/types.rb
peteryates: I don't understand why it would be needed in two places. I just thought this would be a class that copies data from the CSV into a table. 

Maybe we should talk through it tomorrow, I'm sure there's a reason for it but maybe we're overcomplicating it.

peteryates:
Some thoughts.

The original issue #50 said a minimal `GIAS::School` model, I think we've gone a long way beyond minimal here:

* validation - ok some makes sense but this is data we *import* from a place that has already been validated and we don't touch it
* eligibility scopes, checks, etc - these should be in the service layer with the rest of the business logic
* we import the 'raw' data and then use scopes/logic to work things out. Would it make more sense to do that up front as part of the import. For example, instead of `#in_england?` that checks the district code, would having a `country` field be easier?

tonyheadford (approved):

peteryates (approved):

PR #367, opened 2024-09-27
Initial GIAS import

ltello:
This PR contains basic functionality to do the first import of schools and their links from the GIAS api.


package.json
```diff
@@ -7,7 +7,7 @@
     "sass": "^1.79.3"
   },
   "devDependencies": {
```
peteryates, line 9: ✂️

peteryates (approved):
Looks good to me.

I'm not sure we need both `type_code` and `type_name` though, especially as it means we need to hold the mapping too. What happens with this other than eligibility checking?

peteryates:
Also eastings and northings don't make much sense. If we want to do anything geographic we'll need to convert them to [a proper datatype](https://github.com/DFE-Digital/gias-query-tool/blob/main/dml/import_schools.sql#L89) anyway. I think we should drop them for now.

ltello:
> Also eastings and northings don't make much sense. If we want to do anything geographic we'll need to convert them to [a proper datatype](https://github.com/DFE-Digital/gias-query-tool/blob/main/dml/import_schools.sql#L89) anyway. I think we should drop them for now.

We agreed to leave them for now as it is but I can exclude them as well.

ltello:
> Looks good to me.
> 
> I'm not sure we need both `type_code` and `type_name` though, especially as it means we need to hold the mapping too. What happens with this other than eligibility checking?

Ah ok, so you just want to work either with codes or names and do not keep the mapping?.

ltello:
All done!

PR #378, opened 2024-09-29
Induction record processing

tonyheadford:
We've spiked this idea for a way to process induction records from ECF1 and it seems like a good enough approach to build upon. 
This PR adds the basis of the induction record sanitizer and a service to extract school periods from those induction records which could be followed for extracting other period data from a ECF1 induction records.

### DIFF
--- a/app/migration/induction_record_sanitizer.rb
+++ b/app/migration/induction_record_sanitizer.rb
@@ -13,10 +13,11 @@ class InductionRecordSanitizer
   end
 
   def validate!
-    # TODO: I imagine there will be a lot of things needed in here to check/report on
-    does_not_have_multiple_blank_end_dates!
-    does_not_have_multiple_active_induction_statuses!
-    induction_record_dates_are_sequential!
+    validate_induction_records!(induction_records:)
+  end
+
+  def compress_and_validate!
+    validate_induction_records!(induction_records: compress!)
   end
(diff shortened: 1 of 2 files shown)


gracecbrennan:
This is ongoing - has chatted with Pete and Lorenzo on it already

gracecbrennan:
@tonyheadford will be picking this back up today

gracecbrennan:
Going to try to get this working in migration environ.

gracecbrennan:
Getting together with Pete this afternoon to look at sorting things out in migration

claire-hughez:
Tony will catch up on this. He's on support this week so his work on ECF2 will be limited.

claire-hughez:
Tony will try and run the migration in the migration environment and see what happens. Managed to get through all of the support tickets yesterday!!! But already some new ones to deal with in ECF1.

peteryates (approved):
This looks great. It would be really useful as an extension if the output from these comparisons could be viewed (like in that branch I was messing around with a couple of weeks ago that had side-by-side comparisons).

tonyheadford:
> This looks great. It would be really useful as an extension if the output from these comparisons could be viewed (like in that branch I was messing around with a couple of weeks ago that had side-by-side comparisons).

Yes agree

PR #389, opened 2024-10-01
Add teacher migrator

tonyheadford:
Add a migrator to create `Teacher` records using data from the `TeacherProfile` and `User` records in ECF1 that have associated `ParticipantProfile::ECT` and/or `ParticipantProfile::Mentor` records.

Note: this will pull in ~200k records (and that takes about 15mins with this solidqueue config)

**Bonus item** 🎉 
I've added the failures page that lists/displays migration failures instead of having to download them. I've not removed the download functionality in case we find it useful, but it's no longer accessible.
image


app/views/layouts/migration.html.erb
peteryates, line 1: I think we can use `full` here.

peteryates (approved):
Looks good!

PR #404, opened 2024-10-03
Remove require 'rails_helper' from specs

peteryates:
We're already requiring it in .rspec so don't need to do it again 🤦


TobyRet (approved):

ltello (commented):
Cool!

PR #423, opened 2024-10-07
[147] - Create sub navigation view component

TobyRet:
### Ticket
https://github.com/orgs/DFE-Digital/projects/48/views/1?pane=issue&itemId=75955829&issue=DFE-Digital%7Cregister-early-career-teachers%7C147 

### Changes

Creates first iteration of sub navigation view component for the admin page 

### Notes

- Code largely inspired by @peteryates [similar work in the NPQ service](https://github.com/DFE-Digital/npq-registration/pull/1222). The key difference being that we are already using `govuk_service_navigation` view component for the primary navigation, so the sub navigation view component has been made to work independently of this.

### Screenshot

sub_nav

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -1,0 +1,120 @@
+// primary navigation
+//
+// borrowed from x-govuk/govuk-prototype-components
+// https://github.com/x-govuk/govuk-prototype-components/blob/main/x-govuk/components/primary-navigation/_primary-navigation.scss
+.x-govuk-primary-navigation {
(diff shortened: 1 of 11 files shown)


peteryates (approved):
This looks great.

I think the one thing we probably want to tweak is that we decided to go all in on Playwright and haven't used Capybara at all yet.

I have a feeling thought that Capybara is really tightly integrated with ViewComponent's test setup, so there might not be an option.

Probably worth having a chat with the team tomorrow - I'm do love Playwright but I also dislike Capybara enough to want to swim upstream 😅

PR #437, opened 2024-10-09
[430]: Create register ECT flow landing page

TobyRet:
### Ticket
[430](https://github.com/orgs/DFE-Digital/projects/48/views/1?pane=issue&itemId=82590096&issue=DFE-Digital%7Cregister-early-career-teachers%7C430)

### Changes

- create landing page for the register ECT flow
- change 'root' to point to the new landing page
- do not attempt to authenticate the user if they are requesting the landing page

### Screenshot

landing_page


app/controllers/application_controller.rb
```diff
@@ -12,7 +12,7 @@ def current_user
   end
 
   def authenticate
-    return if authenticated?
+    return if authenticated? || request.path == root_path
```
tonyheadford, line 15: Instead of this how do you feel about adding `skip_before_action :authenticate` to the `StartController`?

TobyRet: Ah yes good spot - I'll change

claire-hughez:
Just to be clear, this won't be the landing page - it's the first page of the journey when a user clicks 'Add an early career teacher'

config/routes.rb
```diff
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
-  root to: 'pages#home'
+  root to: "start#index"
```
peteryates, line 2: As @claire-hughez mentioned, this will be visible after the user's logged in and probably linked to from their 'dashboard' in the app.

The path's probably going to be something like:

```
/school/register-ect/start
```

This will eventually be the first page of the wizard flow, but for now it can just belong to a regular controller.

TobyRet: Thanks for clarifying @claire-hughez & @peteryates  - I've updated the PR to move the page to the correct location

tonyheadford (approved):

PR #447, opened 2024-10-10
Fix login screen when personas is disabled

peteryates:
The personas_path is now unreachable when `ENABLE_PERSONAS` is false, so we want to avoid calling the helper elsewhere. Moving the logic to a helper tidies the logic up and makes it easier to test.


claire-hughez:
We did this to stop people without accounts being able to sign in to the migration environment, where in future they might be able to view personal data.

PR #454, opened 2024-10-11
fix: correction to ECT landing page

TobyRet:
### Changes

- Update to body copy
- Make sure page title renders correctly

### DIFF
--- a/app/views/school/register_ect/start.html.erb
+++ b/app/views/school/register_ect/start.html.erb
@@ -1,4 +1,4 @@
-<% page_data(title: sanitize("What you'll need"), error: false) %>
+<% page_data(title: "What you'll need", error: false) %>
 
 <h2 class="govuk-heading-m">Personal information</h2>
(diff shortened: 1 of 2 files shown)


app/views/school/register_ect/start.html.erb
```diff
@@ -1,12 +1,12 @@
-<% page_data(title: "What you'll need", error: false) %>
+<% page_data(title: sanitize("What you'll need"), error: false) %>
```
peteryates, line 1: Might be worth calling `sanitize` from the `_head.html.erb` partial so it's always applied to titles?

claire-hughez (commented):
Works for me once the changes Pete suggests are done :) Thanks Toby!

app/views/school/register_ect/start.html.erb
TobyRet: Yep good spot - have moved it to `_head.html.erb`

peteryates (approved):

PR #466, opened 2024-10-14
chore: deploy the review app if files in 'app' folder are touched

TobyRet:
### Context

Currently the review app is on deployed if the author remembers to add the 'Deploy' label. Sometimes this is forgotten. This is bad.

### Changes

Update the pipeline config to deploy the review app if any of the files in the `app` folder change. Restricting it to this folder will mean we do not create any unnecessary review app deploys (for example when Dependabot updates dependencies)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,4 @@
+Review:
+- changed-files:
+  - any-glob-to-any-file:
+    - app/**
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -47,17 +47,28 @@ jobs:
           docker-repository: ghcr.io/dfe-digital/ecf2
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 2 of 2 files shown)


peteryates:
Might it make sense to add the label if a change was detected in `app`, then allow the workflow to detect it and continue as normal?

I think changing the label name from 'Deploy' to 'Review' would probably make sense too.

claire-hughez:
I would also very much appreciate if the label is renamed to review.

TobyRet:
> Might it make sense to add the label if a change was detected in `app`, then allow the workflow to detect it and continue as normal?
> 
> I think changing the label name from 'Deploy' to 'Review' would probably make sense too.

Yep makes sense! I push the change

peteryates:
I've renamed the label to `Review`

TobyRet:
I'll remove the comment in the controller once you've had a chance to re-review @peteryates 
This was added to trigger the review app deploy

PR #482, opened 2024-10-15
Add pass and fail actions to TRS API client

peteryates:
These make use of the new endpoint `/v3/persons/{trn}/induction` which allows us to set a status. It's hardcoded to respond with 204 for now and available [in preprod](https://preprod.teacher-qualifications-api.education.gov.uk/swagger/index.html).

The valid statues are currently but liable to change:

```
0 = "Exempt"
1 = "Fail"
2 = "FailedinWales"
3 = "InductionExtended"
4 = "InProgress"
5 = "NotYetCompleted"
6 = "Pass"
7 = "PassedinWales"
8 = "RequiredtoComplete"
```


gracecbrennan:
@peteryates will pick up today

PR #485, opened 2024-10-16
feat: view list of ECTs added to school

TobyRet:
### Goal
Create a placeholder page that will provide:

- A starting point for the 'Add ECT' journey
- A way to view the ECTs added in a school, so that we can see when an ECT has successfully been added

### Ticket
https://github.com/DFE-Digital/register-early-career-teachers/issues/468 

### Review app link
https://cpd-ec2-review-485-web.test.teacherservices.cloud/school/home/ects

### Screenshot
Early_career_teachers__ECT__-_Register_early_career_teachers_-_GOV_UK


app/views/school/home/index.html.erb
```diff
@@ -0,0 +1,34 @@
+<% page_data(title: "Early career teachers (ECT)", error: false) %>
+
+<%= govuk_button_link_to("Add an ECT", school_register_ect_start_path) %>
+<hr class="govuk-section-break--m"/>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-one-third">
+    <h2 class="govuk-heading-m">Search by name</h2>
```
TobyRet, line 7: In the prototype the filter comes from MoJ - have left this off for now but can add it in if we are sure it will be used. Will require some additional set up.

app/views/school/home/index.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <% @relationships.map do |relationship| %>
+      <%= govuk_summary_card(title: link_to_ect(relationship[:ect])) do |card|
+        card.with_summary_list(
+          rows: [
+            { key: {
+              text: "Status",
+            },
+              value: {
+                text: render(Status::StatusTag.new(relationship[:status]))
```
TobyRet, line 18: I'm just using a temporary view component here in order to display the status (similar to be ECF1 does) but let me know if this is overkill.

app/controllers/school/home_controller.rb
```diff
@@ -0,0 +1,8 @@
+class School::HomeController < ApplicationController
```
TobyRet, line 1: If there are any general comments or preferences for how things are named, please let me know.

app/services/school/ect_service.rb
```diff
@@ -0,0 +1,21 @@
+class School::ECTService
+  def self.all(school_id)
+    School::ECTService.new.all(school_id)
+  end
+
+  def all(_school_id)
```
TobyRet, line 6: At the moment I am hard-coding the data but there is also the possibility to pull data from seeds. Is now a good time to do this, or too premature?

config/routes.rb
```diff
@@ -59,6 +59,7 @@
   end
 
   namespace :school do
-    get "/register-ect/start", to: "register_ect#start"
+    get "/home/ects", to: "home#index", as: :ects_home
+    get "/register-ect/start", to: "register_ect#start", as: :register_ect_start
```
TobyRet, line 63: Should we be including a school id in this url or hold fire? (the prototype does not do this)

app/controllers/school/home_controller.rb
```diff
@@ -0,0 +1,8 @@
...
+
+  layout "full"
+
+  def index
+    @relationships = School::ECTService.all(schoolId: 1)
```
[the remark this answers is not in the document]
TobyRet: Oops sorry - yes this is probably misleading. We shouldn't be using camelcase and we would get the id from params. Currently the route does not support a school_id param - this is one of the questions I have asked below (whether we introduce this now)

config/routes.rb
peteryates: I don't think so, when someone logs in with DfE Sign-in they're logging in _as_ a school, so we should limit what they see to that school. If they want to switch they can do it via DfE Sign-in and view from another school's perspective.

app/services/school/ect_service.rb
peteryates: Yeah, I think using the seeds from the start is a good idea.

app/helpers/school/home_helper.rb
```diff
@@ -0,0 +1,8 @@
+module School::HomeHelper
+  def link_to_ect(name)
+    govuk_link_to(name,
+                  '#',
+                  no_visited_state: true,
+                  class: ("govuk-!-font-weight-regular"))
```
peteryates, line 6: Wouldn't the weight be regular by default?

app/views/school/home/index.html.erb
peteryates: I think as it's just wrapping the govuk label component it could just be a helper, depends really on how much logic we expect to be here.

app/controllers/school/home_controller.rb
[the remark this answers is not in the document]
peteryates: I think when we have DfE Sign-in working we'll be setting a `@current_school` variable in the `SchoolController` (which this would inherit from).

app/views/layouts/full.html.erb
```diff
@@ -8,6 +8,8 @@
     <%= govuk_skip_link %>
     <%= govuk_header(full_width_border: true, html_attributes: { class: [environment_specific_header_colour_class] }) %>
     <%= govuk_service_navigation(service_name: "Register early career teachers") do |service_navigation| %>
+      <%= service_navigation.with_navigation_item(text: "Your ECT's", href: school_ects_home_path, current: current_page?(school_ects_home_path)  )%>
```
peteryates, line 11:
```suggestion
      <%= service_navigation.with_navigation_item(text: "Your ECTs", href: school_ects_home_path, current: current_page?(school_ects_home_path)  )%>
```

app/views/layouts/full.html.erb
```diff
@@ -8,6 +8,8 @@
...
+      <%= service_navigation.with_navigation_item(text: "Your mentors", href: '#') %>
```
peteryates, line 12: I raised #488 to make these nav options area specific.

app/services/school/ect_service.rb
TobyRet: Cool - made an attempt at this! More complicated than I thought!

app/helpers/school/home_helper.rb
TobyRet: Yep! Fixed

PR #508, opened 2024-10-21
Add migrator to create ECTAtSchoolPeriods from induction records

tonyheadford:
Add a new migrator to create `ECTAtSchoolPeriod` records that associate a `Teacher` with a `School` as an ECT.
This migrator does the following:
 - Fetches `ParticipantProfile::ECT` records from ECF
 - Validates the participant's `InductionRecord` collection
 - Extracts school periods from the `InductionRecord`s
 - Adds `ECTAtSchoolPeriod` records to reflect these periods
 
From my initial testing I expect this is going to take a few hours to migrate. Also I expect this to have quite a few failures based on induction record issues.


app/migration/migrators/ect_at_school_period.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      migrate(self.class.ects.includes(:induction_records, :teacher_profile, teacher_profile: :user)) do |participant_profile|
+        Rails.logger.info("  --> #{participant_profile.id}")
+
+        teacher = ::Teacher.find_by!(trn: participant_profile.teacher_profile.trn)
+
+        induction_records = InductionRecordSanitizer.new(participant_profile:)
+        induction_records.validate!
+
+        school_periods = SchoolPeriodExtractor.new(induction_records:)
```
peteryates, line 36: Not related to this PR but it might be neat if `SchoolPeriodExtractor` returned a set of [data](https://docs.ruby-lang.org/en/3.2/Data.html) or structs, rather than reaching into a hash.

peteryates (approved):
Looks good, left one comment on the development env check.

app/migration/migrators/ect_at_school_period.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def self.dependencies
+      %i[teacher]
+    end
+
+    def self.reset!
+      if Rails.env.development?
+        ActiveRecord::Base.connection.execute("TRUNCATE #{::ECTAtSchoolPeriod.table_name} RESTART IDENTITY CASCADE")
+      end
+    end
```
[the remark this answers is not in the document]
tonyheadford: That is a very good idea

app/migration/migrators/ect_at_school_period.rb
tonyheadford: Yes nice idea

peteryates (approved):
Looks great!

PR #562, opened 2024-10-29
Move error summaries to the top of the page

peteryates:
The desgin system suggests that error summaries are placed at the top of the page, beneath the back/breadcrumbs but before any other content.

> Put the error summary at the top of the main container. If your page includes breadcrumbs or a back link, place it below these, but above the \<h1\>.

This change introduces the error_summary content area which we can pass error summaries to so they're rendered in the right place.

| Before | After |
| ----- | ----- |
| Screenshot from 2024-10-29 10-03-33 | image |


peteryates:
The spacing at the top is a bit off, we appear to be missing the `main` wrapper. I'll fix that separately.

ltello (approved):

PR #567, opened 2024-10-29
Make teacher search work with unaccented strings

### LINKED ISSUE #549: Make the teacher search able to cope with accented characters
We have a entry in the seed data called André but searching for Andre yields no results. It should, most people don't know how to type a é.

We can fix this by using [unaccent](https://www.postgresql.org/docs/current/unaccent.html) for both generating the `search` field and performing the full text search.

peteryates:
The way we're searching uses a `search` field in the teachers table that contains the concatenated first_name, last_name and corrected_name fields in a tsvector (for full text indexing).

Unfortunately, this didn't work with accented characters like 'André'; we'd expect to be able to find that record with 'Andre'.

Initially we wanted to simply use [PostgreSQL's unaccent module](https://www.postgresql.org/docs/current/unaccent.html) to standardise the text. This ultimately didn't work because [unaccent is stable but not immutable](https://www.postgresql.org/message-id/flat/201012021544.oB2FiTn1041521@wwwmaster.postgresql.org#201012021544.oB2FiTn1041521@wwwmaster.postgresql.org).

Instead of taking the easy but slow route of just doing the conversions on the fly, this approach is a little more involved.

It won't make any difference in deployed environments where we always migrate rather than repeatedly drop/create/seed the database like we do in dev.

We can solve the problem by adding a new text search confuguration that applies the unaccent module, and use it to build the tsvector we index on.

Without switching from schema.rb to schema.sql, we can't do this as part of the migration but just need to make sure it's set up when the database is built.

The cleanest way to do this is to add a [rake task that enhances the `db:schema:load`](https://www.rubydoc.info/gems/rake/Rake/Task#enhance-instance_method) task by running the snippet of SQL before it. This ensures what we need is in place before creating the tables which depend on it.

Fixes #549

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,10 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2024_10_17_072716) do
+ActiveRecord::Schema[7.2].define(version: 2024_10_28_105436) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_trgm"
   enable_extension "plpgsql"
+  enable_extension "unaccent"
 
   # Custom types defined in this database.
   # Note that some types may not work with other database engines. Be careful if changing database.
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -125,14 +126,14 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "administrative_district_name"
+    t.integer "local_authority_code", null: false
```
peteryates, line 129: Not sure why the field ordering changed here!

peteryates:
I've tested this locally with:

* `bundle exec db:reset`
* `bundle exec db:{drop,setup}`
* `bundle exec db:{drop,create,schema:load,seed}`

All worked fine.

tonyheadford (commented):
Will it still work if the name entered into the search has accents?

db/schema.rb
peteryates: Oh I must've done a full `db:migrate` from scratch at some point after fixing the TRS capitalisation in 0a4dd66e0f170df79a1ea6ac25aedf76f14d59c8.

peteryates:
@tonyheadford Yep, I've added another test to confirm it does.

PR #569, opened 2024-10-29
Add view of teachers

tonyheadford:
Adding the ability to search for and view Teachers

Added an example of how we could compare records, will need tidying/refactoring into something neater and there's lots of logic in the wrong places to make it work currently.

List view of teachers has the search bar and is paginated with `pagy`
Screenshot 2024-10-29 at 17 35 58

The Teacher view shows migrated data in the first column and source ECF data in the second, and I've attempted to highlight the matched fields.
Screenshot 2024-10-29 at 17 02 21


Gemfile
```diff
@@ -15,6 +15,7 @@ gem "tzinfo-data", platforms: %i[windows jruby]
 gem "govuk-components"
 gem "govuk_design_system_formbuilder"
 gem "govuk_markdown"
+gem "pagy", "~> 9.1"
```
peteryates, line 18: I don't think we need Pagy listed explicitly as it's already pulled in by govuk-components.

config/initializers/pagy.rb
peteryates, line 1: Can we trim this?

peteryates (approved):
Looks good. Going to add a separate issue to move the admin CSS to its own place so we keep things organised.

config/initializers/pagy.rb
tonyheadford: Yes, good spot, missed coming back to that

config/routes.rb
```diff
@@ -54,12 +57,19 @@
...
+    post "reset", to: "migrations#reset", as: :reset
+    get ":model/failures", to: "failures#index", as: :failures
+
+    # resources :migrations, only: %i[index create] do
+    #   get ":model/failures", on: :collection, to: "failures#index", as: :failures
+    #   get "download_report/:model", on: :collection, action: :download_report, as: :download_report
+    #   post "reset", on: :collection, action: :reset, as: :reset
+    # end
+    get "/teachers/:trn", to: "teachers#show", as: :teacher_details
   end
```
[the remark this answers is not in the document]
tonyheadford: They are yes, I was juggling with namespaces/modules a bit

PR #592, opened 2024-11-01
Directly link all induction periods to teachers

peteryates:
Originally we made induction periods dependant on ECT at school periods because before the AB portal work came along schools were intended to report the ECT's AB.

However, as the AB portal work jumped the queue we had to add a second relationship that directly links induction periods with teachers, meaning there were two routes between AB and teacher.

This small change made lots of the code we added more complicated and simple things, like validating that a teacher's induction periods don't overlap, arduous.

So, this change simplifies it by removing the link via ECT at school period. It means later we might need to add more logic elsewhere, like when a teacher leaves a school but has an ongoing induction period, we need to deal with it appropriately.

But, for now, this greatly simplifies things.

## Schema before

```mermaid
erDiagram
    appropriate_bodies {
        timestamp_without_time_zone created_at
        character_varying establishment_id
        integer establishment_number
        bigint id PK
        integer local_authority_code
        character_varying name
        timestamp_without_time_zone updated_at
    }

    induction_periods {
        bigint appropriate_body_id FK
        timestamp_without_time_zone created_at
        bigint ect_at_school_period_id FK
        date finished_on
        bigint id PK
        induction_programme induction_programme
        integer number_of_terms
        daterange range
        date started_on
        bigint teacher_id FK
        timestamp_without_time_zone updated_at
    }

    ect_at_school_periods {
        timestamp_without_time_zone created_at
        date finished_on
        bigint id PK
        daterange range
        bigint school_id FK
        date started_on
        bigint teacher_id FK
        timestamp_without_time_zone updated_at
    }

    schools {
        timestamp_without_time_zone created_at
        bigint id PK
        timestamp_without_time_zone updated_at
        integer urn FK
    }

    teachers {
        character_varying corrected_name
        timestamp_without_time_zone created_at
        character_varying first_name
        bigint id PK
        timestamp_without_time_zone induction_start_date_submitted_to_trs_at
        character_varying last_name
        tsvector search
        character_varying trn
        timestamp_without_time_zone updated_at
    }

    induction_periods }o--|| appropriate_bodies : "appropriate_body_id"
    ect_at_school_periods }o--|| schools : "school_id"
    ect_at_school_periods }o--|| teachers : "teacher_id"
    induction_periods }o--|| ect_at_school_periods : "ect_at_school_period_id"
    induction_periods }o--|| teachers : "teacher_id"
```

## Schema after

```mermaid
erDiagram
    appropriate_bodies {
        timestamp_without_time_zone created_at
        character_varying establishment_id
        integer establishment_number
        bigint id PK
        integer local_authority_code
        character_varying name
        timestamp_without_time_zone updated_at
    }

    induction_periods {
        bigint appropriate_body_id FK
        timestamp_without_time_zone created_at
        date finished_on
        bigint id PK
        induction_programme induction_programme
        integer number_of_terms
        daterange range
        date started_on
        bigint teacher_id FK
        timestamp_without_time_zone updated_at
    }

    ect_at_school_periods {
        timestamp_without_time_zone created_at
        date finished_on
        bigint id PK
        daterange range
        bigint school_id FK
        date started_on
        bigint teacher_id FK
        timestamp_without_time_zone updated_at
    }

    teachers {
        character_varying corrected_name
        timestamp_without_time_zone created_at
        character_varying first_name
        bigint id PK
        timestamp_without_time_zone induction_start_date_submitted_to_trs_at
        character_varying last_name
        tsvector search
        character_varying trn
        timestamp_without_time_zone updated_at
    }

    schools {
        timestamp_without_time_zone created_at
        bigint id PK
        timestamp_without_time_zone updated_at
        integer urn FK
    }

    induction_periods }o--|| appropriate_bodies : "appropriate_body_id"
    ect_at_school_periods }o--|| schools : "school_id"
    ect_at_school_periods }o--|| teachers : "teacher_id"
    induction_periods }o--|| teachers : "teacher_id"
```


claire-hughez:
I don't see this as an issue. We'll continue to occasionally have to create induction records for teachers before they are registered by a school, and we won't know their school at that point.

I also like this because it makes clearer the induction record is owned by the appropriate body, which it is for now.

Can we update our diagrams to reflect this?

It's potentially worth thinking through how we'll submit info about induction that schools submit, longer term, such as the programme type or who the appropriate body is.

Potentially, given we know who the AB is or who the LP and DP is might be expressed tentatively to other bodies as an expression of interest, we could have a similar way of categorising and looking after those fields that have dual ownership and might change.

ltello:
A few first thoughts come to my mind after reading this PR:

- To me associating `induction_periods` to a `teacher` instead of to a`ECTAtSchool` sounded like the natural thing to do given that ABs can register an induction (or even a teacher I guess) before a school does.

- I have been disconnected from ECF2 data model since a while ago but the other day I saw this:
```
 a teacher has_many :induction_periods_reported_by_appropriate_body and
 a teacher has_many :induction_periods
```
 and i thought: "Really??. Do we need to query 2 tables in a coordinate manner to get induction info for a teacher?.  Keeping consistency between these 2 sources of induction information will add so much complexity...!"

- When a teacher changes schools do we really need to create a new `induction_period` if the AB keeps being the same for both?. Conceptually it feels like one `induction_period` per teacher per AB unless we store in it information specific to the school the teacher belongs to (which I think is not the case so far).

- Thinking about a school dashboard listing all its participants, the benefit of having `an ect_at_school has many induction_periods` instead of `an ect_at_school has many induction_periods through teacher` is minimum because in both cases for a given `ect_at_school` we would still need to pick the data from a specific (based on conditions about a range of dates) `induction_period`  from a potential set  (which makes queries more complex and less performant) and not from a single straight one (which would make queries simpler and faster).

- Surely this change will make other queries more complex (for instance, those that need to get info from the school of the teacher from a specific `induction_period`). However I think those could be just a few and I think their complexity are nothing compared to maintaining 2 different sources of induction information with all the problems stated above.

peteryates:
> (which makes queries more complex and less performant) and not from a single straight one (which would make queries simpler and faster).

I'm not sure this is the case.  We can easily query based on time ranges. We can still find all the induction periods for a teacher that overlap with any ECT at school periods in a straightforward query.

tonyheadford (approved):
LGTM and makes sense for the AB data

PR #605, opened 2024-11-04
[473] - Register ECT Flow - fetch TRS data

TobyRet:
### Ticket

https://github.com/DFE-Digital/register-early-career-teachers/issues/473 

### Changes

- Find an ECT record in TRS that matches a TRN and date of birth entered by a school user. 
- Display the name associated with the TRS record so that the user can confirm the right person has been found.
- Add magic 'credentials' so that the team can get proceed through the find ECT step on dev environment

### Video

https://github.com/user-attachments/assets/4c7fa656-37ba-413d-9570-53dea8e05afb


app/controllers/schools/register_ect_controller.rb
```diff
@@ -54,5 +76,9 @@ def step_params
     def default_params
       ActionController::Parameters.new({ current_step => params })
     end
+
+    def format_to_govuk_date(date_as_string)
```
TobyRet, line 80: Move to view helper?

app/services/schools/trs.rb
```diff
@@ -0,0 +1,57 @@
...
+      new(trn:, date_of_birth:).fetch_trs_teacher
+    end
+
+    def initialize(trn:, date_of_birth:)
+      @trn = trn
+      @date_of_birth = date_of_birth
+    end
+
+    def fetch_trs_teacher
+      if magic_credentials_match? && !Rails.env.production?
```
TobyRet, line 16: Keen to hear feedback on whether there are better ideas for handling magic credentials

tonyheadford: All deployed envs are "production" so this won't work outside of local dev and specs?

TobyRet: mmm ... ok my bad. Was thinking we also have a 'sandbox' or 'staging' environment like ECF1.
In that case seems like the only option here would be define env vars in the github workflow (per job)?

https://stackoverflow.com/questions/63467662/github-actions-define-workflow-level-environment-variable-between-jobs

tonyheadford: Yes we might want an env var like `ENABLE_MAGIC_CREDENTIALS` and set it to true in the environments we want it.  I've recently done a similar thing with personas and enabling migration testing. There are per-environment files under here `config/terraform/application/config` where you can set this up.  We've also added then to a settings object to make them easier/cleaner to get at `config/initializers/app_settings.rb`

TobyRet: Awesome thanks @tonyheadford - I make the change

peteryates: Also I don't think we should mix fake data/responses with the normal application code wherever possible.

We set up a [fake API client](https://github.com/DFE-Digital/register-early-career-teachers/blob/main/spec/support/api/trs/fake_api_client.rb) for testing. Maybe the lowest impact way to write this would be to check for the presence of an environment variable (`USE_FAKE_TRS_API`) in the `#api_client` method below and if it is use that.

Then the app can be used/deployed either fully real or fully fake, but no confusing middle grounds.

app/components/schools/magic_credentials_guidance_component.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    </span>
+  </summary>
+  <div class="govuk-details__text">
+    <p class="govuk-body">To successfully locate an ECT from the TRS API, use the following information:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>Teacher Reference Number (TRN): 1234567</li>
+      <li>Date of birth: 1/12/2000</li>
+    </ul>
+  </div>
+</details>
```
peteryates, line 14: I don't think this should be styled in a GOV.UK fashion, it's not part of the app non-team member should see and just gives us clues on how to progress. Can we just put it in a yellow box or something?

app/services/schools/trs.rb
TobyRet: That might be a good option. But does this also introduce a potential issue - if we only allow 'real' **or** 'fake' api calls in any environment (ie not both) we may not be able to detect any issues relating to 'real' TRS API calls in the lower environments. Is this an acceptable risk?

peteryates: Yeah, it's a risk but I think when the magic dates are enabled devs are unlikely to use 'real' TRN/DOb combos anyway. I think having the list of working combinations on the page is the best way to approach it, then we don't block progress _and_ we keep testing it.

A toggle for enabling fake ones might be useful locally/dev though, and we can set it to true by default so people who are getting the app running locally have something to work with before they get keys.

app/components/schools/magic_credentials_guidance_component.html.erb
TobyRet: Hey @peteryates  - happy to change but you may want to agree with Claire as Jaz in the first instance as this is what they asked for

claire-hughez: Hey,

I don't feel strongly on this either way and IMO it's very low priority. I just meant it could have a blue background or tag of some kind to differentiate it - I think what is here is fine, but we'll be reusing this throughout for different review points so just something simple and consistent is what is important.

It was blue in the wireframe we agreed so was going with that, it was just in a blue box as in here:

image

TobyRet: Updated guidance component @claire-hughez 

guidance_update

TobyRet: Background colour has been lightened

guidance_updated

app/components/schools/magic_credentials_guidance_component.rb
```diff
@@ -0,0 +1,7 @@
+module Schools
+  class MagicCredentialsGuidanceComponent < ViewComponent::Base
+    def render?
+      ActiveModel::Type::Boolean.new.cast(ENV.fetch('SHOW_MAGIC_CREDENTIALS_GUIDANCE', false))
```
peteryates, line 4: I wonder if this env var name should be more general so it's usable for other helpfil hints we might want to add.

`TEST_GUIDANCE` maybe? Or `ENABLE_TEST_HINTS`?

peteryates: Similarly for the CSS class names etc, especially as they're no longer really magic.

app/components/schools/magic_credentials_guidance_component.html.erb
```diff
@@ -0,0 +1,55 @@
+<div class="magic-credential-guidance">
+  <h3 class="govuk-heading-m">Information to review this journey</h3>
+  <p class="govuk-body">To successfully locate an ECT from the TRS API, use credentials from the table below:</p>
+  <table class="govuk-table">
```
peteryates, line 4: I'd recommend using the [table component](https://govuk-components.netlify.app/components/table/#input-slim-building-a-table-from-arrays) here.

app/views/schools/register_ect/review_ect_details.html.erb
```diff
@@ -1 +1,41 @@
-<% page_data(title: "Check if this is your ECT", error: true, backlink_href: schools_register_ect_find_ect_path ) %>
+<% page_data(title: "Check if this is your ECT", error: @wizard.current_step.errors.present?, backlink_href: schools_register_ect_find_ect_path) %>
+
+<dl class="govuk-summary-list">
```
peteryates, line 3: Worth using the [summary list component](https://govuk-components.netlify.app/components/summary-list/) here.

spec/components/schools/magic_credentials_guidance_component_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+  end
+
+  def with_env_var(key, value)
+    original_value = ENV[key]
+    ENV[key] = value
+    yield
+  ensure
+    ENV[key] = original_value
+  end
```
peteryates, line 31: Might be safer to use `allow(ENV).to receive(:[]).with(key).and_return(true)` here, [like this](https://github.com/DFE-Digital/early-careers-framework/blob/2bab735b17d59eb0853890b0c3244e2281135078/spec/requests/healthcheck_spec.rb#L24).

peteryates (commented):
Looking good! Left a few really minor comments.

app/components/schools/magic_credentials_guidance_component.rb
TobyRet: I've renamed to 'test guidance' throughout (component, class names, env var ...)

spec/components/schools/magic_credentials_guidance_component_spec.rb
TobyRet: I'm not sure if it's less safe as such (as the original value is restored) but happy to follow convention

app/views/schools/register_ect/review_ect_details.html.erb
TobyRet: Done!

peteryates (approved):
Looks good!

Issue #607, opened 2024-11-05
Increase number of instances on Azure


gracecbrennan:
## Why?
We are currently restricted on how much Azure resource we can use due to our plan being on the cheapest number of instances in a few cases. This means, when we attempt migration of data, it quickly becomes very slow. For example, it took ten hours to complete half of the teacher migration

## What?
Grace to explore options for upping the number of instances we have.

peteryates:
This ticket is closely related to #523 - probably makes sense for them to be done together.

claire-hughez:
This is done.

Issue #526, opened 2024-10-23
High availability


peteryates:
In `config/terraform/application/config/production.tfvars.json` set:

```
"postgres_enable_high_availability": true
```

PR #619, opened 2024-11-05
Bump production and migration database plans and increase migration workers

### LINKED ISSUE #523: Database plan
ECF1 has the following PostgreSQL skus set up in production:

```tf
"postgres_flexible_server_sku": "GP_Standard_D4ds_v4",
``` 

We should aim for the same in prod in ECF2 as when it's is fully launched it could be under similar load, although between the AB portal launch in January and the full ECF2 launch after October 2025, its load will be significantly less.

 It currently defaults to `B_Standard_B1ms`.

## Some questions for @DFE-Digital/teacher-services-infrastructure

1. would it make sense for us to launch ECF1 with a lesser plan (maybe `GP_Standard_D2ds_v4`) and then upgrade to `GP_Standard_D4ds_v4` closer to the full ECF2 launch - i.e., is it a big job?
2. what's the easiest way to do this for ECF2 at the moment, none of our services are live, if we just switch the values and deploy, will the small databases be replaced with beefier ones?

peteryates:
See the individual commits for more of an explanation, but this is mainly adjusting the production database setup and:

* switching production from `B_Standard_B1ms` to `GP_Standard_D4ds_v4`
* enabling high availabilty mode
* enabling backup storage

We need the extra horsepower in the migration env because we're soon going to be running full migrations which are incredibly slow with the current setup.

In production our schedule changed. We originally intended to launch the service in late 2025/early 2026but now we need to deliver the Appropriate Body portal in Jan/Feb 2025, so we want to ensure our database properly configured with HA, backups etc. 

Fixes #523, #526, #607

### DIFF
--- a/config/terraform/application/config/migration.tfvars.json
+++ b/config/terraform/application/config/migration.tfvars.json
@@ -2,5 +2,8 @@
     "cluster": "production",
     "namespace": "cpd-production",
     "environment": "migration",
-    "enable_logit": true
+    "enable_logit": true,
+    "worker_replicas": 6,
+    "worker_memory_max": "2Gi",
+    "postgres_flexible_server_sku": "GP_Standard_D4ds_v4"
 }
--- a/config/terraform/application/config/migration.tfvars.json
+++ b/config/terraform/application/config/migration.tfvars.json
@@ -5,5 +5,5 @@
     "enable_logit": true,
     "worker_replicas": 6,
     "worker_memory_max": "2Gi",
-    "postgres_flexible_server_sku": "GP_Standard_D4ds_v4"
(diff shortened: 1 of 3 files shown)


config/terraform/application/config/migration.tfvars.json
```diff
@@ -2,5 +2,8 @@
     "cluster": "production",
     "namespace": "cpd-production",
     "environment": "migration",
-    "enable_logit": true
+    "enable_logit": true,
+    "worker_replicas": 6,
+    "worker_memory_max": "2Gi",
+    "postgres_flexible_server_sku": "GP_Standard_D4ds_v4"
```
peteryates, line 8: Probably worth jumping to v5 here.

PR #639, opened 2024-11-07
[475]: Add 'email' and 'check answers' steps to Register ECT flow

TobyRet:
### Ticket
https://github.com/DFE-Digital/register-early-career-teachers/issues/475 

### Changes

- Add 'enter email' and 'check answers' steps to the Register ECT flow
- Create an ECTPresenter (view model) to access session data and make this accessible to the views that need to show the user previously stored data

### Video 

https://github.com/user-attachments/assets/6ebd5e2e-8249-4caf-b324-4b16bd9fbc6a


app/presenters/schools/ect_presenter.rb
```diff
@@ -0,0 +1,50 @@
...
+        instance_variable_set("@#{key}", value)
+        # attr_readers are dynamically generated with the pattern
+        # "#{wizard_step}_#{attribute}" e.g find_ect_trn
+        self.class.attr_reader(key) unless self.class.method_defined?(key)
+        @dynamic_attr_readers << key
+      end
+
+      return if @dynamic_attr_readers.empty?
+
+      self.class.send(:private, *@dynamic_attr_readers)
```
tonyheadford, line 18: Although this is quite concise I do think it's slightly terrifying.  We are trying hard to avoid "magic" code to make it simple for everyone to understand, work with and debug. Worried we have methods in here calling methods that will only exist once a page has been visited or a journey has been travelled.
Personally I'd probably prefer this to be more explicit and have methods for the data we're collecting. Also possibly make this `TeacherPresenter` so we could reuse for the mentor journeys maybe?  What do you think?

tonyheadford (commented):
Looks great overall, just some comments/thought around the presenter approach

app/presenters/schools/ect_presenter.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def full_name
+      "#{find_ect_first_name} #{find_ect_last_name}"
+    end
```
peteryates, line 23: We can [use `Teachers::Name` for this](https://github.com/DFE-Digital/register-early-career-teachers/blob/main/app/services/teachers/name.rb).

spec/features/schools/register_an_ect_spec.rb
```diff
@@ -65,4 +71,29 @@ def and_i_should_see_the_ect_details_in_the_review_page
...
+    page.click('text=Confirm and continue')
+  end
+
+  def then_i_should_be_taken_to_the_email_address_page
+    path = '/schools/email-address'
+    expect(page.url).to end_with(path)
+  end
+
+  def when_i_enter_the_ect_email_address
+    page.fill('#email-address-email-field', 'example@example.com')
```
peteryates, line 85: Maybe use the label as a selector here?

```
page.get_by_label(label).fill(number.to_s)
```

TobyRet: Possibly, but the danger here is that the label is more likely to change and could make the tests more brittle?

tonyheadford (approved):
Looks great, thanks for refactoring the presenter, looks much better

app/presenters/schools/ect_presenter.rb
peteryates: Ah yeah - fair point.

spec/features/schools/register_an_ect_spec.rb
peteryates: I think given the balance between readability at the expense of brittleness, I'd go with the former.

Plus, if the labels are long I think you can just look for part of it with a regexp, which should reduce the brittleness significantly:

```ruby
page.get_by_label(/TRN/).fill('1234567')
```

PR #650, opened 2024-11-08
Make persona-based AB signin smoother

peteryates:
Once we're integrated with DfE Sign-in we'll store the current appropriate_body_id for AB users in the session. When using personas for development and testing purposes, we don't really have a clean way of linking them.

This change adds the AB name to the seeds, and if the user signs in with a persona and that persona name contains the appropriate body's name it sets the session correctly.

It could be optimised further using a database query but we'll only have a handful of seed data so this should do for now.


tonyheadford (approved):
LGTM!

PR #684, opened 2024-11-12
Make the test guidance component more flexible

peteryates:
This PR makes the following small changes:

* adds `TEST_GUIDANCE=true` to `.env.example`
* moves the test guidance component to a shared location
* makes the test guidance component a bit more flexible by allowing custom content to be passed in

## Before

```
<%= render TestGuidanceComponent.new %>
```

### After

```ruby
<%=
  render TestGuidanceComponent.new do |component|
    component.with_trs_example_teacher_details
  end 
%>
```

or

```ruby
<%=
  render TestGuidanceComponent.new do
    "some other content entirely"
  end 
%>
```


tonyheadford (approved):
Nice!

PR #693, opened 2024-11-13
Start new version of ab service rules

peteryates:
This PR makes moves the old version of the appropriate bodies service rules to a file with a `.v1.md` extension and adds a fresh one that reflects what's in the register early career teachers app.

We decided to do this because it's difficult to tell which:

* rules were purely ECF1 related
* features have been implemented in register early career teachers
* are still valid/necessary from an register early career teachers point of view

## Reviewer notes

It'll be much easier to review commit by commit, the first file move change creates an unnecessary diff.

## Preview

| New homepage with old/new service rules side by side | Keyless layout |
| --------- | ----------- |
| image| image |


documentation/service-rules/appropriate-body-users.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Ongoing induction with another appropriate body
+
+If the early career teacher has already been registered by an appropriate body, they cannot be registered by another.
+
+The first appropriate body must [release them](#releasing-an-early-career-teacher) before the second can register them.
+
+#### Prohibited from teaching
+
+Teachers who have been prohibited from teaching aren't permitted to receive receive induction.
```
peteryates, line 112: I couldn't find anything in the policy that explicitly backs this up. I know it's common sense but it would be nice to link to something.

claire-hughez:
If we can avoid versioning and just use RECT for new stuff, and TRA/Manage training for old stuff, that would be swell, but no worries if its too tricky

documentation/service-rules/appropriate-body-users.md
claire-hughez: I think in order to be teaching you need to not be prohibited from teaching.

And in order to serve teaching induction, you need to be teaching.

So it kinda follows on from there?

PR #697, opened 2024-11-14
Add training period migrator

tonyheadford:
Adds training periods migration to `Teacher` migrator.  This creates `TrainingPeriod` records by interrogating teachers `InductionRecord`s to determine the periods and providers involved.
This will fail if there is no `ECTAtSchoolPeriod` or `MentorAtSchoolPeriod` that encapsulates the training period dates.
For this first iteration I'm only bringing in FIP training records.


peteryates (approved):
Looks like a great start!

PR #698, opened 2024-11-14
WIP

ltello:


app/models/user.rb
```diff
@@ -1,8 +1,16 @@
...
   validates :name, presence: true
   validates :email, presence: true, uniqueness: true, notify_email: true
 
   # TODO: we can encrypt the secret in the DB but we need to set up keys first
   # encrypts :otp_secret
+
+  # Associations
   has_many :dfe_roles
+
+  # Instance Methods
```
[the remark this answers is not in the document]
tonyheadford: Yes agree, generally if it isn't a `FIXME:`, `TODO:` or a `NOTE:` comment we probably don't need them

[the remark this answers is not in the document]
ltello: I like to have small model classes with similar structure in their definition and methods ordered alphabetically.
If we commit to it we can get rid of it but sometimes it is hard to find things inside a classes because there is no order at all.

app/models/school.rb
```diff
@@ -9,7 +9,19 @@ class School < ApplicationRecord
...
+
   # Instance Methods
+  delegate :address_line1, :address_line2, :address_line3, to: :gias_school
+  delegate :administrative_district_name, :closed_on, :establishment_number, to: :gias_school
+  delegate :funding_eligibility, :induction_eligibility, :in_england, to: :gias_school
+  delegate :local_authority_code, :local_authority_name, to: :gias_school
+  delegate :name, :primary_contact_email, to: :gias_school
+  delegate :opened_on, :phase_name, :postcode, :type_name, to: :gias_school
+  delegate :secondary_contact_email, :section_41_approved, :status, to: :gias_school
+  delegate :ukprn, :website, to: :gias_school
```
tonyheadford, line 23: Could we use the GIAS School for the school data and not `delegate`?
Personally I don't like delegate in database models as I feel it obsfucates relationships and the sources of data and you can't use them in queries. It would be better to use a presenter type of pattern when really necessary and keep the models clean and simple IMO.

app/models/school.rb
```diff
@@ -9,7 +9,19 @@ class School < ApplicationRecord
             presence: true,
             uniqueness: true
 
+  # Scopes
+  scope :search, ->(q) { includes(:gias_school).references(:gias_schools).where("gias_schools.search @@ websearch_to_tsquery('unaccented', ?)", q) }
```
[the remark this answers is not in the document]
ltello: Well... we work with School instances all the time not GIAS::School ones.
Maybe we can twick it but I think it is ok like this.

app/models/school.rb
ltello: Actually I have the opposite view in the School model.

We almost always work with School in the service instead of GIAS::Schools. 
If a School instance is not able to respond to messages about their basic properties (name, status, urn... etc), what is the purpose of having the School class?

On the other hand, delegation allows encapsulation of a School entity so the rest of the service doesn't need to know about the internals on how their properties are stores, where, etc, preventing things like school.gias_school.name in favour of school.name according to the Law of Demeter.

peteryates: I'm normally against delegation (and always against it when [prefixes are involved](https://www.rubydoc.info/docs/rails/Module:delegate)) but here I think it's ok as the school model is a kind of a empty 'pointer' at the true GIAS school.

I think the above snippet could be neater with a single list alphabetically sorted delegations though:

```suggestion
  delegate
    :address_line1,
    :address_line2,
    :address_line3
    :administrative_district_name,
    :closed_on,
    :establishment_number,
    :etc,
    to: :gias_school
```

app/models/user.rb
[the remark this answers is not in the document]
tonyheadford: I think if we follow the principle of keeping the models as clean an small as possible there should be no need for the comments and we can check that things are grouped together in PRs?

app/models/school.rb
tonyheadford: True this is an unusual case, but I'd like us to avoid it in DB models as a general rule.

app/models/user.rb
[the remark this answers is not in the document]
peteryates: We do kind of already have these comments/groupings across lots of the other models. I don't really mind them but no strong feelings either way.

peteryates (approved):
Nice, looks like a good first step.

I think there's a few bits that could be tidied up as per comments.

db/migrate/20241113090244_add_search_column_to_gias_schools.rb
```diff
@@ -0,0 +1,10 @@
+class AddSearchColumnToGIASSchools < ActiveRecord::Migration[7.2]
+  def change
+    add_column :gias_schools, :search, :tsvector, type: :tsvector, as: "to_tsvector('unaccented', #{search_columns})", stored: true
+    add_index :gias_schools, :search, using: :gin
+  end
+
+  def search_columns
+    %w[name postcode urn].map { |col| "coalesce(#{col}::text, '')" }.join(" || ' ' || ")
```
peteryates, line 8: I wonder about URN actually - in the teachers search I omit it from the `search` column and scan the string for 7 digit numbers, and search for them directly.

app/models/user.rb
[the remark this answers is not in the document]
ltello: Well, probably more important than the comments themselves, I would like us to commit to order methods alphabetically in the classes so it is easy to navigate and find them.

app/services/schools/teachers.rb
peteryates, line 1: It does feel a little bit like this is preempting things that we might not actually need, or at least yet.

app/services/schools/teachers.rb
```diff
@@ -0,0 +1,53 @@
...
+                                       .select("DISTINCT ON (teacher_id) *")
+    end
+
+    def latest_period_of_all_mentors
+      @latest_period_of_all_mentors ||= @school
+                                       .mentor_at_school_periods
+                                       .includes(:teacher)
+                                       .order("teacher_id, started_on DESC")
+                                       .select("DISTINCT ON (teacher_id) *")
+    end
```
peteryates, line 51: Why would we need the latest period for all ECTs/mentors? 🤔

db/migrate/20241113090244_add_search_column_to_gias_schools.rb
ltello: I saw it for Teachers but why not leave the nice Postgresql search engine do the job for us instead?

app/services/schools/teachers.rb
ltello: These are the base queries to calculate the participants to display to a school in the dashboard.
I mean, the actual Teacher instances are obtained from these 2 methods.

Maybe I can make them private for now.
If we need to access them publicly in the future (perhaps to display the date a participant joined a school, or maybe the school email of the participant...) we can move them to the public interface of the class.

peteryates: But we'd only want the ongoing periods at this point and there's an existing scope for that in interval.

My fear is that these methods won't actually get used. I'd rather start small and add when we know we need them.

app/services/schools/teachers.rb
ltello: I just added the `current_*` versions to make it clear the distinction between the two groups and make the developer think which one is the correct one for the feature they are developing in the future. 
But happy to remove it because it is not of any use yet.

app/services/schools/teachers.rb
ltello: Why only the ongoing?

I think in ECF1 we don't only list ongoing participants in the admin interface but also completed ones, the ones that left the school... I would say, we are listing all of them.

Kind of the same in the school dashboard

tonyheadford: I don't think any of this has been designed tested in UR yet @ltello so we don't know if we'll need any of the same approaches in the UI as we had in ECFv1.  I'd agree with @peteryates and start small and we can add this type of thing when it becomes a need.

app/views/admin/schools/index.html.erb
```diff
@@ -0,0 +1,42 @@
+<% page_data(title: "Schools") %>
+
+<div class="search-box">
+  <%= form_with method: :get, class: "search-box__form" do |f| %>
+    <div class="search-box__input">
+      <%= f.govuk_text_field(
+            "q",
+            value: @q,
+            label: { text: "Search for school", size: "s" },
+            hint: { text: "Name or URN or Postcode or Contact email address" },
```
claire-hughez, line 10:
```suggestion
            hint: { text: "Name, URN or postcode" },
```

I think we descoped contact email address, so just edited here

app/views/admin/schools/show.html.erb
```diff
@@ -0,0 +1,173 @@
...
+end %>
+
+<h2 class="govuk-heading-m">Mentors</h2>
+
+<%= govuk_table do |table|
+  table.with_head do |head|
+    head.with_row do |row|
+      row.with_cell(text: "Name")
+      row.with_cell(text: "Corrected Name")
+      row.with_cell(text: "TRN")
```
claire-hughez, line 158: @tonyheadford @peteryates @ltello keen for other thoughts, but I hadn't imagine we'd always show corrected name and name in admin tooling like this. I thought we'd probably just show the corrected name if given, or the name otherwise. It could be an additional field when you click in.

If it's helpful for you guys for now, fine to leave, but given this is mainly for migration purposes, I'm guessing it won't be. Unless we plan to check the DQT API for the name given there when we migrate records.

claire-hughez (commented):
Looks great, thanks Lorenzo. I've suggested 1 minor content tweak and asked a question in the chat we might want to think about.

app/views/admin/schools/show.html.erb
ltello: I think displaying the 2 names is helpful when solving Support requests from schools.
But happy to leave just one if you want.

claire-hughez: I think let's remove to just name for now, with the logic, give corrected name if it exists, otherwise give name. If that's okay please.

The vast majority of records shouldn't have corrected names, so it feels unnecessary here, but we should add it when we create a page for individual ECTs being viewed by admins. Thanks for responding so quickly.

Issue #512, opened 2024-10-22
Enter an ECT's national insurance number to match a TRS record


claire-hughez:
## Description
This story is part of the 'Add an ECT' journey.

When an ECT's record cannot be found by searching for the TRN and DOB in TRS, we should ask them for the ECT's national insurance number (NINO). This might help find the record and we would then search in TRS for a match using TRN and National insurance number. If the record still can't be found, the user will not be able to continue the journey.

## User story

**As** a school user
**I need** to find an ECT's record in DQT
**So that** I can validate their teaching record is eligible for funding and training

## Prototype, screenshots & content
This is the NINO page if we've not been able to find on TRN / DOB: [Prototype link](https://ecf2-prototy-v3-oz6554dztlkwtv.herokuapp.com/v3/school/nino)

If still not found, then we have this page: [Prototype link](https://ecf2-prototy-v3-oz6554dztlkwtv.herokuapp.com/v3/school/not-found)

From this [page in the prototype](https://ecf2-prototy-v3-oz6554dztlkwtv.herokuapp.com/v3/school/find-ect) you enter a TRN as `9999999`, it then redirects to the NINO page. If you enter a NINO starting with `Z`, then it will redirect to the not found page.

Screenshots of the relevant pages:
![image](https://github.com/user-attachments/assets/57571f34-b87f-4eb7-b678-4c508966ea7a)

Content should be as per the prototype.
URLs and validation should be as per [this doc](https://educationgovuk.sharepoint.com/:x:/r/sites/TeacherServices/Shared%20Documents/Teacher%20Continuing%20Professional%20Development/Teacher%20CPD%20Team/Register%20early%20career%20teachers/Beta/Design/Register%20ECTs%20page%20titles%20and%20URLs.xlsx?d=w0c1578ea643147b29357c7f610ee178f&csf=1&web=1&e=U8iIdi).

## Acceptance criteria

- [x] 1. When user submits a valid TRN and date of birth on the 'Find an ECT' page, if the check for a TRS record on TRN and date of birth combined fails to find an ECT's record, user is navigated to the 'We cannot find your ECT's details' page.
- [x] 2. On the 'We cannot find your ECT's details' page:
    - [x] 2a. User is asked to give a National insurance number for that ECT.
[...]

PR #699, opened 2024-11-14
Enter an ECT's national insurance number to match a TRS record

TobyRet:
### Ticket

#512 

### Changes 

- Add a 'national insurance number' step to the Register ECT flow and direct the user to this step if TRS API cannot find a teacher with the submitted date of birth and TRN 
- Wire up national insurance number validator
- Direct the user to a 'not found' page if the TRS API can still not find a teacher with the submitted national insurance number
- Add national insurance numbers to test guidance component
- Refactoring: create a factory to handle storage for each step in the flow

#### Enter national insurance number

https://github.com/user-attachments/assets/9b407cfd-21f7-4645-b1a1-131e9ad57ba5

#### Teacher not found

https://github.com/user-attachments/assets/c11881d2-6107-410b-aab2-1154367d7f46


app/services/appropriate_bodies/claim_an_ect/find_ect.rb
```diff
@@ -17,7 +17,11 @@ def import_from_trs!
         #       below a case and add different errors to the :base
         return unless pending_induction_submission.valid?(:find_ect)
 
-        pending_induction_submission.assign_attributes(appropriate_body:, **trs_teacher.present)
+        pending_induction_submission
+          .assign_attributes(
+            appropriate_body:,
+            **trs_teacher.present.except(:trs_national_insurance_number)
```
TobyRet, line 23: @peteryates - I have also started calling the #present method in the register ECT code, however I need the national insurance number whereas pending inductions does not. I'm not sure how coupled pending inductions will need to be to the #present method in future but if you prefer, I can avoid using this and making changes to suit my use case.

PR #719, opened 2024-11-18
Fix a bug in migrator

tonyheadford:
The `TrainingPeriod` builders use `OpenStruct` to find the correct school period to which they belong. However this means we need to `require "ostruct"` before using it.


peteryates:
If we know what we're putting in them can it not just be a struct or [data](https://docs.ruby-lang.org/en/3.2/Data.html)?

PR #755, opened 2024-11-20
Updated AB title from register to record

RoseAsaf:
We want to change this because the word register appears twice so record feels more apt. 

image


peteryates:
Note the extra `title: ...` thing will be fixed by #757

claire-hughez:
looks cool! shouldn't it be a plural inductionS?

peteryates:
Going to merge for now to keep things moving but we can revisit pluralising.

PR #756, opened 2024-11-20
[2105] Custom domains

saliceti:
## Description
Configuration for custom domains: infrastructure as well as staging, sandbox and production environments.

The zone has been created. Waiting for DNS delegation.


saliceti:
[EC2: create domains](https://trello.com/c/tKYSyWTo/2105-ec2-create-domains)

config/terraform/domains/infrastructure/terraform.tf
```diff
@@ -0,0 +1,19 @@
+terraform {
+  required_version = "= 1.6.4"
```
peteryates, line 2: We're on `1.9.8` [elsewhere](https://github.com/DFE-Digital/register-early-career-teachers/blob/7c2e6b9e66137c97b4f56813cd401b28bc1b49db/config/terraform/application/terraform.tf#L4).

saliceti: Updated

peteryates (approved):

PR #767, opened 2024-11-21
Refactoring

ltello:
This PR refactors a bit the code of a journey.
Easier to review it looking at each commit and their purpose.

The main goal is to:
- move the logic behind a step into their own step class as opposed to external handler classes.
- move the logic to determine the next step inside each step

Other minor changes perhaps required by the main goals are:
- store only stringify keys in the wizard store because deeper levels of data structures are stringified by default by the session.
- simplify the data persisted in a wizard data store by not grouping them by step so it is easier to be accessed only by key as opposed to (step_name, key)
- allow TRS requests with trn only


app/wizards/schools/register_ect/base_wizard.rb
```diff
@@ -8,12 +8,13 @@ class BaseWizard < DfE::Wizard::Base
       steps do
         [
           {
-            find_ect: FindECTStep,
-            review_ect_details: ReviewECTDetailsStep,
-            email_address: EmailAddressStep,
             check_answers: CheckAnswersStep,
             confirmation: ConfirmationStep,
+            email_address: EmailAddressStep,
```
TobyRet, line 13: Another option could be to order them as per the actual flow (maybe this would need a comment to explain) - may help with debugging or head-scratching.

If not perhaps we could had a doc / diagram at some point to map out the flow visually

app/wizards/schools/register_ect/find_ect_step.rb
```diff
@@ -13,8 +13,45 @@ def self.permitted_params
       end
 
       def next_step
+        return :not_found unless stored_trs_teacher
+        return :national_insurance_number unless birth_date_matches?
+
         :review_ect_details
       end
+
+      def perform
```
TobyRet, line 22: Given this `perform` method is no longer being accessed from a StoreHandler maybe we should rename for greater clarity? e.g `store_in_session`, `save_to_session` or something more descriptive

app/wizards/schools/register_ect/stored_step.rb
```diff
@@ -8,27 +8,36 @@ class StoredStep < DfE::Wizard::Step
...
         return false unless valid_step?
 
-        handler = StepStoreHandlerFactory.create!(
-          step_name: wizard.current_step_name,
-          wizard:,
-          store:,
-          key:
-        )
-        handler.perform
+        perform
```
TobyRet, line 17: See https://github.com/DFE-Digital/register-early-career-teachers/pull/767/files#r1852524058

app/wizards/schools/register_ect/stored_step.rb
```diff
@@ -8,27 +8,36 @@ class StoredStep < DfE::Wizard::Step
...
         store
       end
 
+      def self.permitted_params
+        []
+      end
+
     private
 
       def key
```
TobyRet, line 30: I think this method can be safely removed?

TobyRet (approved):
Great job @ltello - thanks for doing this.
Just a couple of minor comments from me.

app/wizards/schools/register_ect/base_wizard.rb
ltello: I think this is just a list of steps for the Dfe-Wizard gem to know what step class to initialize on each step of the journey.
What I mean is that an ordered list is enough for us to check we have/haven't listed them all here.
We won't be checking this list on our day to day, so I think alphabetic order is ok?
Also the order of execution is not linear, steps can go earlier and later when we play with the back buttons and the  user changes their answers in the check-list

app/wizards/schools/register_ect/find_ect_step.rb
ltello: Happy to rename this method. I wouldn't like to mention the session though in the name but the store or the concept of persisting data. 
To me, the session is this concept is more the current implementation we have for the store (which could be changed in the future to the database for instance instead of the session?)

app/wizards/schools/register_ect/stored_step.rb
ltello: I'll check.

TobyRet (approved):

app/wizards/schools/register_ect/find_ect_step.rb
```diff
@@ -13,8 +13,47 @@ def self.permitted_params
...
+        store.set("date_of_birth", format_date(date_of_birth))
+        store.set("trs_teacher", trs_teacher.presence&.present)
+      end
+
+    private
+
+      def birth_date_matches?
+        return false unless trs_date_of_birth && provided_date_of_birth
+
+        Date.parse(trs_date_of_birth) == Date.parse(provided_date_of_birth)
```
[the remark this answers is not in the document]
peteryates: It is now.

#792

gracecbrennan:
Ready to be merge but there is an issue in the review app that is preventing ability to release the lock. 

To check with Pete/Tony.

~/.cache/zsh/compcache/brew_all_commands
peteryates, line 1: I don't think this should be here

~/.cache/zsh/compcache/brew_casks
peteryates, line 1: Or this

peteryates:
Can we split the Terraform upgrade off to a separate PR?

ltello:
Yes, sure, I was just trying to see if this fixes the issue

peteryates:
I removed the lockfile, deploying should work once the lint error is fixed 🤞🏽

ltello:
Thank you @peteryates.

PR #773, opened 2024-11-21
Register ECT Flow - fix tests 

TobyRet:
### Changes

- Rename feature spec files (they were not running before but now they are!)
- Updates paths in feature specs


TobyRet:
Closing as @ltello has raised the same fix

PR #775, opened 2024-11-22
made induction plural in the title

RoseAsaf:
induction > inductions


peteryates (approved):
Looks good. I think singular makes more sense though.

PR #778, opened 2024-11-22
feat: auto-populate Find ECT form

TobyRet:
### Changes

Add a button to each row in the TestGuidance component. When clicked this will populate the Find ECT form with the TRN and date of birth (or national insurance number)

### Notes
This is using JS but this component would never be shown in prod so ... ok?

#### Happy path

https://github.com/user-attachments/assets/ef6171ef-f29a-41c7-857b-b59fb1c12deb

#### National Insurance step

https://github.com/user-attachments/assets/f4943aa1-8493-41f0-8bed-54ee3583b73a

### DIFF
--- a/app/components/test_guidance_component.rb
+++ b/app/components/test_guidance_component.rb
@@ -26,11 +26,25 @@ class TestGuidanceComponent < ViewComponent::Base
       ]
     end
 
+    def rows_with_buttons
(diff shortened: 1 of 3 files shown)


peteryates:
This is great, definitely makes it easier to skip through the form.

peteryates (approved):

PR #790, opened 2024-11-22
Limit the records appropriate bodies can see

peteryates:
**Note this follows on from #789, review that first**

This change locks down the queries used in appropriate body controllers so they can only see what they're meant to see, namely:

* teachers who have an ongoing induction period with them
* pending induction submissions that belong to them
* induction periods that belong to them

I suggest reviewing it commit-by-commit.


gracecbrennan:
@edujackedu to review

Issue #758, opened 2024-11-20
Assign a new mentor to an ECT when there are no mentors currently active at a school


mason-emily:
## Description
This story delivers the functionality to assign a new mentor to an ECT at a school where there are no mentors already currently active at the school. 

Note, there will be separate tickets for different routes if there are mentors already active at the school.

## User story

**As** school user
**I need** to assign a mentor to an ECT at my school
**So that** the mentor is reported to DfE so that they can gain access to the ECT's materials and in some cases access to mentor training 

## Prototype, screenshots & content
[Link to prototype](https://ecf2-prototy-v4-k7mv50fp1jl0ue.herokuapp.com/v4/school/home/ects#)

Screenshots of relevant screens: 
Note, the parts crossed out in red are in the prototype but are **not in scope** for this ticket.
![Image](https://github.com/user-attachments/assets/8c5b5bc0-afdf-4725-9511-0c5508ed9347)

Content should be as per the prototype.
URLs should be as per [this doc](https://educationgovuk.sharepoint.com/:x:/r/sites/TeacherServices/Shared%20Documents/Teacher%20Continuing%20Professional%20Development/Teacher%20CPD%20Team/Register%20early%20career%20teachers/Beta/Design/Register%20ECTs%20page%20titles%20and%20URLs.xlsx?d=w0c1578ea643147b29357c7f610ee178f&csf=1&web=1&e=U8iIdi).

## Acceptance criteria

- [x] 1. On the <ins>List of ECTs at a school</ins> page, when there **is** a mentor assigned, the ECT's summary card currently shows the mentor's name. When there **is no mentor assigned** for an ECT, the 'Mentor' line in the ECT's summary card should show a notice saying 'You must assign a mentor or register a new one for this ECT' as per the prototype.
- [x] 2. In this notice, the text 'assign a mentor or register a new one' will be a hyperlink to the <ins>'What you'll need to add a new mentor for [ECT name]'</ins> page.
- [x] 3. The <ins>'What you'll need to add a new mentor for [ECT name]'</ins> page will include content on the details needed to register a mentor as per the prototype.
- [x] 4. There will be a 'Back' link on the <ins>'What you'll need to add a new mentor for [ECT name]'</ins> page which navigates the user back to the <ins>List of ECTs at a school</ins> page.

## Technical notes
Seed data already ready for this.

claire-hughez:
Nearly there, have to add a couple of tests and will then submit the PR to review.

PR #805, opened 2024-11-25
[#758]: assign a new mentor to an ect when there are no mentors currently active at a school

ltello:
[Ticket](https://github.com/DFE-Digital/register-early-career-teachers/issues/758)

Video:

https://github.com/user-attachments/assets/0237cf65-eb84-4c70-8faa-97457dc6d680


app/controllers/schools/home_controller.rb
```diff
@@ -1,18 +1,37 @@
...
 
-    def school
+    def set_school
       # This is temporary. 'School' will be set once DfE signin hooked up
-      @school ||= School.joins(:gias_school)
-                        .select("schools.id, gias_schools.name")
-                        .first
+      # School in the session or first school with ects but no mentors or first school
+      @school = (school_from_session || first_school_with_no_mentors_but_ects || first_school).tap do |school|
+        session[:school_id] = school&.id
```
tonyheadford, line 17: We should be using URN for the school identifier

app/controllers/schools/register_mentor_controller.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Schools
+  class RegisterMentorController < ApplicationController
+    before_action :initialize_wizard, only: %i[new create]
+    before_action :reset_wizard, only: :new
+
+    FORM_KEY = :register_mentor_wizard
+    # WIZARD_CLASS = Schools::RegisterMentor::Wizard.freeze
```
tonyheadford, line 9: Where is WIZARD_CLASS defined?

ltello: This is just almost a copy of the register-ect-controller.rb. 
I just kept it like this because it is coming to play in the next PR

Issue #687, opened 2024-11-12
ECT being registered already has an induction completion date


mason-emily:
## Description
This story delivers the functionality to check for an induction completion date associated with the ECT's TRS record when we find a match. ECTs who already have an induction completion date should not be registered in the service as they have already completed induction, so we end the registration journey in this case.

## User story

**As** a school user
**I need** to be told if the ECT I'm registering has already completed induction
**So that** I know why I cannot register the ECT

## Prototype, screenshots & content
[Prototype link to the 'You cannot register [ECT name]' page](https://ecf2-prototy-v4-k7mv50fp1jl0ue.herokuapp.com/v4/other-pages/add-ect/induction-completed)

Screenshots of relevant screens for if the ECT ****does have an induction completion date** on their TRS record:
![image](https://github.com/user-attachments/assets/8d0904bd-46db-4867-bb11-5ec2eb77a392)

Content should be as per the prototype.
URL should be as per [this doc](https://educationgovuk.sharepoint.com/:x:/r/sites/TeacherServices/Shared%20Documents/Teacher%20Continuing%20Professional%20Development/Teacher%20CPD%20Team/Register%20early%20career%20teachers/Beta/Design/Register%20ECTs%20page%20titles%20and%20URLs.xlsx?d=w0c1578ea643147b29357c7f610ee178f&csf=1&web=1&e=U8iIdi).

## Acceptance criteria

- [x] 1. When user submits a valid TRN and date of birth on the <ins>'Find an ECT'</ins> page and a match is found for the ECT's TRS record using both TRN and date of birth:
    - [x] 1a. If the ECT's record **has** an induction completion date, user will be navigated to the <ins>'You cannot register [ECT name]'</ins> page. 
    - [x] 1b. If the ECT's record **does not have** an induction completion date, user will continue to be navigated to the <ins>'Check if this is your ECT'</ins> page.
- [x] 2. On the <ins>'You cannot register [ECT name]'</ins> page:
    - [x] 2a. H1 text will include the ECT's first and last name from their TRS record 
    - [x] 2b. Content will be as per the prototype - note, this is slightly different to #688 .
    - [x] 2c. There will be a link to 'Register another ECT' which navigates the user to the <ins>'Find an ECT'<ins> page where the fields should be reset to blank.
    - [x] 2d. There will be no 'Back' link on the <ins>'You cannot register [ECT name]'</ins> page.
[...]

PR #807, opened 2024-11-25
End register ECT journey if teacher has completed their induction

TobyRet:
### Ticket 

#687 

### Changes

- Update TestGuidanceComponent: add 'Notes' column so that additional information about a teacher can be included.
- If a teacher has completed their induction, the user is redirected to a new error page 

### Video

https://github.com/user-attachments/assets/0fd12031-c790-402f-8f00-db38843b1fcb


app/views/schools/register_ect/induction_completed.html.erb
```diff
@@ -0,0 +1,5 @@
+<% page_data(title: "You cannot register #{@ect.full_name}", error: false) %>
+
+<p class="govuk-body">Our records show <%= @ect.full_name %> has already completed their induction and is not eligible for further ECT training.</p>
+
+<p class="govuk-body"><%= govuk_link_to("Register another ECT", schools_register_ect_find_ect_path) %></p>
```
claire-hughez, line 5:
```suggestion
<p class="govuk-body">Our records show <%= @ect.full_name %> has already completed their induction and is not eligible for further training as an ECT.</p>

<p class="govuk-body"><%= govuk_link_to("Register another ECT", schools_register_ect_find_ect_path) %></p>
```

tonyheadford: Could this be something like: "... has already completed their induction and therefore cannot be registered as an ECT." ?

tonyheadford (approved):
Looks fine from the dev perspective

app/views/schools/register_ect/induction_completed.html.erb
claire-hughez: That works for me! What do you think Rob?

Issue #688, opened 2024-11-12
ECT being registered has an induction status of 'exempt'


mason-emily:
## Description
This story delivers the functionality to check for an induction status of 'exempt' associated with the ECT's TRS record when we find a match. ECTs who have an induction status of exempt should not be registered in the service as they are exempt from completing induction, so we end the registration journey in this case.

## User story

**As** a school user
**I need** to be told if the ECT I'm registering is exempt from induction
**So that** I know why I cannot register the ECT

## Prototype, screenshots & content
[Prototype link to the 'You cannot register [ECT name]' page](https://ecf2-prototy-v4-k7mv50fp1jl0ue.herokuapp.com/v4/other-pages/add-ect/induction-status-exempt)

Screenshots of relevant screens for if the ECT ****does have an induction status of exempt** on their TRS record:
![image](https://github.com/user-attachments/assets/6d99c0eb-f4dc-4e36-8287-94bd31331c9e)

Content should be as per the prototype.
URL should be as per [this doc](https://educationgovuk.sharepoint.com/:x:/r/sites/TeacherServices/Shared%20Documents/Teacher%20Continuing%20Professional%20Development/Teacher%20CPD%20Team/Register%20early%20career%20teachers/Beta/Design/Register%20ECTs%20page%20titles%20and%20URLs.xlsx?d=w0c1578ea643147b29357c7f610ee178f&csf=1&web=1&e=U8iIdi).

## Acceptance criteria

- [x] 1. When user submits a valid TRN and date of birth on the <ins>'Find an ECT'</ins> page and a match is found for the ECT's TRS record using both TRN and date of birth:
    - [x] 1a. If the ECT's record **has** an induction status of 'exempt', user will be navigated to the <ins>'You cannot register [ECT name]'</ins> page. 
    - [x] 1b. If the ECT's record **does not have** an induction status of 'exempt', user will continue to be navigated to the <ins>'Check if this is your ECT'</ins> page.
- [x] 2. On the <ins>'You cannot register [ECT name]'</ins> page:
    - [x] 2a. H1 text will include the ECT's first and last name from their TRS record 
    - [x] 2b. Content will be as per the prototype - note, this is slightly different to #687.
    - [x] 2c. There will be a link to 'Register another ECT' which navigates the user to the <ins>'Find an ECT'<ins> page where the fields should be reset to blank.
    - [x] 2d. There will be no 'Back' link on the <ins>'You cannot register [ECT name]'</ins> page.
[...]

PR #816, opened 2024-11-26
End ‘register ECT’ journey if teacher is exempt from induction

TobyRet:
### Ticket
#688 

### Changes
- Add test data for 'exempt' teacher in TestGuidance component
- If a teacher is exempt from induction, the user is redirected to a new error page

### Video 

https://github.com/user-attachments/assets/928506d7-8d96-4f66-a807-b037f2b3ab8a

### DIFF
--- a/app/views/schools/register_ect/induction_exempt.html.erb
+++ b/app/views/schools/register_ect/induction_exempt.html.erb
@@ -1,5 +1,5 @@
 <% page_data(title: "You cannot register #{@ect.full_name}", error: false) %>
 
-<p class="govuk-body">Our records show <%= @ect.full_name %> is exempt from having to serve induction so you cannot register them for ECF training.</p>
+<p class="govuk-body">Our records show <%= @ect.full_name %> is exempt from having to serve induction and therefore cannot be registered as an ECT.</p>
 
 <p class="govuk-body"><%= govuk_link_to("Register another ECT", schools_register_ect_find_ect_path) %></p>
--- a/app/views/schools/register_ect/induction_exempt.html.erb
+++ b/app/views/schools/register_ect/induction_exempt.html.erb
@@ -1,0 +1,5 @@
+<% page_data(title: "You cannot register #{@ect.full_name}", error: false) %>
+
+<p class="govuk-body">Our records show <%= @ect.full_name %> is exempt from having to serve induction so you cannot register them for ECF training.</p>
+
+<p class="govuk-body"><%= govuk_link_to("Register another ECT", schools_register_ect_find_ect_path) %></p>
--- a/app/wizards/schools/register_ect/find_ect_step.rb
+++ b/app/wizards/schools/register_ect/find_ect_step.rb
@@ -31,8 +31,7 @@ module Schools
                    trs_trn: trs_teacher.trn,
                    trs_first_name: trs_teacher.first_name,
                    trs_last_name: trs_teacher.last_name,
-                   trs_induction_completed?: trs_teacher.induction_status == 'Pass',
-                   trs_induction_exempt?: trs_teacher.induction_status == 'Exempt')
+                   trs_induction_status: trs_teacher.induction_status)
       end
 
       def trs_teacher
--- a/app/wizards/schools/register_ect/find_ect_step.rb
+++ b/app/wizards/schools/register_ect/find_ect_step.rb
@@ -14,9 +14,9 @@ module Schools
 
       def next_step
         return :trn_not_found unless ect.in_trs?
-        return :induction_completed if ect.induction_completed? && ect.matches_trs_dob?
-        return :induction_exempt if ect.induction_exempt? && ect.matches_trs_dob?
         return :national_insurance_number unless ect.matches_trs_dob?
+        return :induction_completed if ect.induction_completed?
+        return :induction_exempt if ect.induction_exempt?
 
         :review_ect_details
       end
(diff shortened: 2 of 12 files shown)


app/wizards/schools/register_ect/find_ect_step.rb
```diff
@@ -15,6 +15,7 @@ def self.permitted_params
       def next_step
         return :trn_not_found unless ect.in_trs?
         return :induction_completed if ect.induction_completed? && ect.matches_trs_dob?
+        return :induction_exempt if ect.induction_exempt? && ect.matches_trs_dob?
         return :national_insurance_number unless ect.matches_trs_dob?
 
         :review_ect_details
```
ltello, line 21: Wouldn't be better to reorder sentences like this:

```
def next_step
  return :trn_not_found unless ect.in_trs?
  return :national_insurance_number unless ect.matches_trs_dob?
  return :induction_completed if ect.induction_completed?
  return :induction_exempt if ect.induction_exempt? 

  :review_ect_details
end
```

app/wizards/schools/register_ect/find_ect_step.rb
```diff
@@ -30,7 +31,8 @@ def persist
                    trs_trn: trs_teacher.trn,
                    trs_first_name: trs_teacher.first_name,
                    trs_last_name: trs_teacher.last_name,
-                   trs_induction_completed?: trs_teacher.induction_status == 'Pass')
+                   trs_induction_completed?: trs_teacher.induction_status == 'Pass',
+                   trs_induction_exempt?: trs_teacher.induction_status == 'Exempt')
```
ltello, line 35: Wouldn't be better to just save `trs_induction_status` and put the logic for 
`trs_induction_completed?: trs_teacher.induction_status == 'Pass'` and 
`trs_induction_exempt?: trs_teacher.induction_status == 'Exempt'`
in Schools::RegisterECT::ECT class?

app/wizards/schools/register_ect/national_insurance_number_step.rb
```diff
@@ -14,6 +14,7 @@ def self.permitted_params
       def next_step
         return :not_found unless ect.in_trs?
         return :induction_completed if ect.induction_completed? && ect.matches_trs_national_insurance_number?
+        return :induction_exempt if ect.induction_exempt? && ect.matches_trs_national_insurance_number?
```
ltello, line 17: I think we don't need to check for `&& ect.matches_trs_national_insurance_number?` in these 2 sentences because if `ect.in_trs?` is true is because TRS returned a record that matches both the TRN and the NINO.
So at this point `ect.matches_trs_national_insurance_number?` will always be true.
Actually we can even delete this method `matches_trs_national_insurance_number?`

app/wizards/schools/register_ect/national_insurance_number_step.rb
```diff
@@ -26,7 +27,8 @@ def persist
                    trs_date_of_birth: trs_teacher.date_of_birth,
                    trs_first_name: trs_teacher.first_name,
                    trs_last_name: trs_teacher.last_name,
-                   trs_induction_completed?: trs_teacher.induction_status == 'Pass')
+                   trs_induction_completed?: trs_teacher.induction_status == 'Pass',
+                   trs_induction_exempt?: trs_teacher.induction_status == 'Exempt')
       end
```
ltello, line 32: Same as above

config/routes.rb
```diff
@@ -104,6 +104,8 @@
 
       get "induction-completed", action: :new
```
ltello, line 106: maybe we can remove this blank line for the routes file not to become huge and less readable?

ltello (changes requested):
Good job Toby, just a few suggestions

app/views/schools/register_ect/induction_exempt.html.erb
```diff
@@ -0,0 +1,5 @@
+<% page_data(title: "You cannot register #{@ect.full_name}", error: false) %>
+
+<p class="govuk-body">Our records show <%= @ect.full_name %> is exempt from having to serve induction so you cannot register them for ECF training.</p>
```
claire-hughez, line 3:
```suggestion
<p class="govuk-body">Our records show <%= @ect.full_name %> is exempt from having to serve induction and therefore cannot be registered as an ECT.</p>
```

TobyRet: Cool - have made the change to this page and the 'induction completed' error page also

app/wizards/schools/register_ect/find_ect_step.rb
TobyRet: Yes good point. I have moved the logic.

app/wizards/schools/register_ect/national_insurance_number_step.rb
TobyRet: Good spot. I've removed. At this point we have already queried the TRS API with the trn and NINO so if we get something back we do not need to check again.

app/wizards/schools/register_ect/find_ect_step.rb
ltello: This ☝️

ltello (changes requested):
Just one change left I guess

app/wizards/schools/register_ect/find_ect_step.rb
TobyRet: 557b40de279667c4a6b6771d8dc9b657b06e8587

Issue #764, opened 2024-11-21
Find a mentor to add for an ECT at your school using TRN and NiNo


mason-emily:
## Description
This story will provide the functionality for school users to find the mentor they are trying to add using a national insurance number (combined with TRN) when date of birth doesn't match a TRS record. 

## User story

**As** school user
**I need** to find a mentor's record in TRS
**So that** I can continue the journey of registering the right mentor and assigning them to an ECT at my school

## Prototype, screenshots & content
[Prototype mentor journey
](https://ecf2-prototy-v4-k7mv50fp1jl0ue.herokuapp.com/v4/school/mentor/what-youll-need)

[Prototype asking for National Insurance number](https://ecf2-prototy-v4-k7mv50fp1jl0ue.herokuapp.com/v4/school/mentor/nino)

**Screenshots of relevant screens:**
Note, the 'corrected name' question is in the prototype but not in scope for this ticket - will be added later.
![image](https://github.com/user-attachments/assets/25e1602c-8dc9-413b-bdef-41d8412c0293)

Content should be as per the prototype.
URL and validation messages should be as per [this doc](https://educationgovuk.sharepoint.com/:x:/r/sites/TeacherServices/Shared%20Documents/Teacher%20Continuing%20Professional%20Development/Teacher%20CPD%20Team/Register%20early%20career%20teachers/Beta/Design/Register%20ECTs%20page%20titles%20and%20URLs.xlsx?d=w0c1578ea643147b29357c7f610ee178f&csf=1&web=1&e=U8iIdi).

## Acceptance criteria

- [x] 1. When user submits a valid TRN and date of birth on the <ins>'Find a mentor'</ins> page, if the check for a TRS record on TRN and date of birth combined fails to find a mentor's record, user is navigated to the <ins>'We cannot find the mentor's details'</ins> page.
- [x] 2. On the <ins>'We cannot find the mentor's details'</ins> page:
    - [x] 2a. User is asked to give a National insurance number for that mentor.
[...]

PR #840, opened 2024-11-28
[764]: find a mentor to add for an ect at your school using TRN and NINo

ltello:
[Issue #764](https://github.com/DFE-Digital/register-early-career-teachers/issues/764)

Video: 

https://github.com/user-attachments/assets/598f035d-f648-4369-93f7-fa9cc3d8bc2b

### DIFF
--- a/spec/views/schools/register_mentor/national_insurance_number.html.erb_spec.rb
+++ b/spec/views/schools/register_mentor/national_insurance_number.html.erb_spec.rb
@@ -1,0 +1,43 @@
+RSpec.describe "schools/register_mentor/national_insurance_number.html.erb" do
+  let(:back_path) { schools_register_mentor_find_mentor_path }
+  let(:continue_path) { schools_register_mentor_national_insurance_number_path }
+  let(:title) { "We cannot find the mentor's details" }
+  let(:wizard) { Schools::RegisterMentor::Wizard.new(current_step: :national_insurance_number, store: {}) }
+
+  before do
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "sass": "^1.81.0"
   },
   "devDependencies": {
-    "playwright": "^1.49.0",
+    "playwright": "^1.48.2",
```
TobyRet, line 10: Interesting - should be a separate commit with explanation?

spec/views/schools/register_mentor/national_insurance_number.html.erb_spec.rb
```diff
@@ -0,0 +1,49 @@
+RSpec.describe "schools/register_mentor/national_insurance_number.html.erb" do
```
TobyRet, line 1: I need to add these view specs to find_ect

TobyRet (commented):
One comment about the change to package.json otherwise LGTM

package.json
ltello: I get this as the latest compatible package version. Can you run bin/setup and see what version it installs for you?

peteryates: Still, if possible we should keep anything to do with versions in separate PRs. Makes it far easier should we need to roll anything back.

ltello: Fine, removed from this PR

TobyRet (approved):

PR #852, opened 2024-12-01
Fix guidance select buttons on the AB claim flow

peteryates:
Various small tweaks and changes:

- Move JS files from app/javascript to app/assets/javascripts and apply linter
- Fix capitalisation of TRN field (make it sentence case)
- Move JS include to the test guidance component
- Ensure there's a pending induction submission (so the early exit pages are renderable)


peteryates:
The contrast's still not great on links.

image

I saw this coming [a while ago](https://github.com/DFE-Digital/register-early-career-teachers/pull/605#discussion_r1829081055). Maybe we should use another colour? Yellow? Or Pink?

Screenshot From 2024-12-01 13-36-16

peteryates:
I'm merging this now so it's working for UR shortly.

claire-hughez:
> The contrast's still not great on links.

This is my fault mainly for insisting on blue (the best colour). Happy for you guys to change to whatever passes contrast.

Issue #804, opened 2024-11-25
Write initial ECT data to DB


mason-emily:
## Description
This ticket is to persist the data from the 'Add an ECT' journey to the DB

At the successful end of the journey once the "Check your answers" step has been accepted we should persist the data we have collected.

Once the transfer and training parts of the journey have been tested and agreed we can add an additional tickets to handle those cases, but for now creating a `Teacher` and associating them correctly with the `School` is enough.

Note, we're not including ECT email address as part of this ticket as we first need to explore how we'll handle this data #642 

## Acceptance criteria

- [ ] 1. Create and populate a new `Teacher` record with the name and TRN details entered
- [ ] 2. Create an `ECTAtSchoolPeriod` record to associated the `Teacher` with the `School` that is being administered. This should have the appropriate `starting_on` date.

## Linked tickets
#478

claire-hughez:
Will merge personas one, and then will clean up the PR for this. Will then get some feedback from Lorenzo.

gracecbrennan:
Going to bring @tonyheadford in on discussion on this

TobyRet:
I've set up a meeting with AB and ECF devs on Friday morning to discuss this PR and general approaches we may want to align on as add to the code base.

### Meeting summary

_I’ve raised a PR related to writing data to the DB in the Register ECT flow, [which has led to some interesting discussions](https://github.com/DFE-Digital/register-early-career-teachers/pull/859#discussion_r1865897901) and some bigger questions about use of wizards and how we implement the designs.

Given similar work may be happening in AB-land, Tony and I thought might be a good to take stock and have a broader discussion about the development work done so far in Register ECT and Mentors and AB. Specifically, we wondered if it may be useful to align on high-level approaches across apps —such as patterns, wizards, name-spacing, and the use of service objects, to name a few.

While we’re still in the early stages and there’s limited value in aligning on every detail just yet, but there could be value in identifying commonalities, pain points, and whether we can agree on some key principles moving forward._

gracecbrennan:
Discussion ongoing about what we do with this.

gracecbrennan:
Reraised a PR for this

PR #859, opened 2024-12-02
Write session data to the database (register ECT journey)

TobyRet:
### Ticket
#804 

### Changes

- Create ECT in the database
- (lots of renaming) - adding '_wizard' to folders and routes to make is consisent with recent changes to Register Mentor Flow

### Notes
Follows the 'write to db pattern' in the Register Mentor process following lengthy discussion ...


app/wizards/schools/register_ect/ect.rb
```diff
@@ -1,6 +1,14 @@
 module Schools
   module RegisterECT
-    class ECT < SimpleDelegator
+    class ECT
```
TobyRet, line 3: Removing `< SimpleDelegator` has it's downsides - the class is way more verbose. However the reason for doing this was so that we could be consistent and inject the `Schools::Teacher` service (in order to save the ECT to the db). Is there a better way?

app/wizards/schools/register_ect/wizard.rb
```diff
@@ -30,7 +30,7 @@ def self.step?(step_name)
       delegate :reset, to: :ect
 
       def ect
-        @ect ||= ECT.new(store)
+        @ect ||= ECT.new(session_repository: store, teacher_repository:)
```
TobyRet, line 33: It's annoying that the DfE wizard gem only caters for one type of storage, but I believe this may change in future versions.

app/services/schools/teacher.rb
```diff
@@ -18,6 +18,18 @@ def fetch_etcs_and_mentors
...
+    def create!(trs_first_name:, trs_last_name:, trn:)
+      ActiveRecord::Base.transaction do
+        teacher = ::Teacher.create!(first_name: trs_first_name, last_name: trs_last_name, trn:)
+
+        ECTAtSchoolPeriod.create(
+          teacher:,
+          school:,
+          started_on: 1.month.ago
+        )
+      end
```
ltello, line 30: I expected this method to be in `Schools::RegisterECT::ECT` instead.

app/controllers/schools/register_ect_controller.rb
```diff
@@ -29,7 +29,8 @@ def initialize_wizard
       @wizard = WIZARD_CLASS.new(
         current_step:,
         step_params: params,
-        store: SessionRepository.new(session:, form_key: FORM_KEY)
+        store: SessionRepository.new(session:, form_key: FORM_KEY),
+        teacher_repository: Schools::Teacher.new(school.id)
```
ltello, line 33: I don't see why we need `Schools::Teacher` in the wizard. We have already `Schools::RegisterECT::ECT`

app/wizards/schools/register_ect/ect.rb
ltello: If we move the creation of the real `Teacher` record  and `ECTsAtSchoolPeriod` associated into this class (as suggested above), we don't need to remove the `SimpleDelegator` inheritance.

TobyRet: I think the 'SimpleDelegator` is great in how it simplifies things but I think a downside is that is not clear in this instance what we are delegating to (ie it is not obvious that the ECT class talks to the SessionRepository). For a person who is unfamiliar with the code, they would need to find where the ECT class is initialised in order to determine this relationship. 

If we write to the directly to the db in the ECT class as well then I feel we are breaking SRP and are not working with dependencies at the same level of abstraction. 

This was the reason why in the controller I initialised the ECT class as follows:

```
        @ect ||= ECT.new(session_repository: store, teacher_repository:)
```

The ECT class is now responsible for communicating to 'repositories' and nothing else.

By removing the 'SimpleDelegator' inheritance in the ECT class we can clearly see what is being injected into the class. But the downside as I mentioned is that the class is more verbose, but, I do not see this as a evil thing as such.

app/controllers/schools/register_ect_controller.rb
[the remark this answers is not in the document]
tonyheadford: Yes that makes sense

app/wizards/schools/register_ect/ect.rb
tonyheadford: IMO I think we should have a distinct service to create a teacher and maybe separate ones to associate the teacher with the school - i.e. adds the ECTAtSchoolPeriod or MentorAtSchoolPeriod. I reckon we might be creating a rod for our backs with the `Schools::Teacher` and `Teacher::` namespaces in the services and we definitely don't want to be doing the same thing in both places, so maybe some thought up front as to what type of services should live in which namespace because there will inevitably be some cross-over.  Perhaps the ECT/Mentor related stuff should be in `Schools::` and teacher under `Teachers::` but also where will training and induction related services live?
I think this should be a service object outside of any wizard or UI related code that can be invoked as needed from anywhere.
e.g.
```
ActiveRecord::Base.transaction do
  teacher = Teacher::Create.new(details from wizard)
  period = Schools::ECT::Create.new(teacher:, other details from wizard)
  ... other stuff
end
```
Hope I've understood this correctly 😅

ltello (approved):

documentation/site/lib/filters/govuk_markdown.rb
peteryates, line 1: This file is from [the documentation site](https://ecf-service-manual.education.gov.uk/) rather than the Rails app. Not sure it needs to be here.

peteryates (approved):
Looks good to me

PR #861, opened 2024-12-02
Added release and outcomes rules to doc

joe-harrison-dfe:


documentation/service-rules/appropriate-body-users.md
```diff
@@ -118,4 +118,47 @@ Teachers who have been prohibited from teaching aren't permitted to receive rece
 
 ## Releasing an early career teacher
 
+When an ECT finishes an induction at an appropriate body before they have fully completed their induction, the appropriate body inform us by releasing them.
+
+The release process takes place with the following steps:
+
+1. Locate the ECT's record from the 'view list of ECT records' page using the following information:
+    * TRN
+    * Date of Birth
```
claire-hughez, line 127: Hm, is this true? To me this kinda sounds like they have to give TRN and date of birth to release them. They just need to find the record from the View page, right?

documentation/service-rules/appropriate-body-users.md
```diff
@@ -118,4 +118,47 @@ Teachers who have been prohibited from teaching aren't permitted to receive rece
...
+    * name
+    * date of birth
+    * email address
+    * [QTS](https://www.gov.uk/guidance/qualified-teacher-status-qts) award date
+    * QTS status
+    * induction start date
+    * induction status
+    * [initial teacher training](https://www.gov.uk/government/collections/initial-teacher-training) provider
+    * initial teacher training end date
+    * [alerts](https://www.gov.uk/government/collections/teacher-misconduct)
```
claire-hughez, line 138: Is this bit relevant to releasing or the view page more generally? Is the important bit here that they click the 'Release' button on the record?

documentation/service-rules/appropriate-body-users.md
```diff
@@ -118,4 +118,47 @@ Teachers who have been prohibited from teaching aren't permitted to receive rece
...
+* **TRN**
+  - must be present
+  - must be 7 numeric digits
+* **Date of birth**
+  - must be present
+  - the teacher must be between 18 and 100 years old
+* **Induction start date**
+  - must be present
+  - must be after the QTS award date
+* **induction period end date**
```
claire-hughez, line 154:
```suggestion
* **Induction period end date**
```

just to be consistent with capitalisation 🤓

documentation/service-rules/appropriate-body-users.md
```diff
@@ -118,4 +118,47 @@ Teachers who have been prohibited from teaching aren't permitted to receive rece
...
+  - must be present
+  - must be later than the induction start adte
```
claire-hughez, line 156:
```suggestion
  - must be later than the induction start date
```

documentation/service-rules/appropriate-body-users.md
```diff
@@ -118,4 +118,47 @@ Teachers who have been prohibited from teaching aren't permitted to receive rece
...
+    * induction status
+    * [initial teacher training](https://www.gov.uk/government/collections/initial-teacher-training) provider
+    * initial teacher training end date
+    * [alerts](https://www.gov.uk/government/collections/teacher-misconduct)
+3. Release the ECT by entering the following information:
+    * The end date of the induction period
+    * The number of terms carried out during the induction period
+
+### Validation
+
```
claire-hughez, line 144: hm, I'm not sure about this section - in the release flow they wouldn't be validating quite a few of these. Should this be a Header 2 (with 2 hashes) for validation rules for the entire service? If so, it would be helpful to then have each section as a Header 3 (with 3 hashes). I've left an example of what I mean :)

documentation/service-rules/appropriate-body-users.md
```diff
@@ -118,4 +118,47 @@ Teachers who have been prohibited from teaching aren't permitted to receive rece
...
+* **TRN**
+  - must be present
+  - must be 7 numeric digits
+* **Date of birth**
+  - must be present
+  - the teacher must be between 18 and 100 years old
```
claire-hughez, line 150:
```suggestion
## Validation rules for the service

### TRN

  - must be present
  - must be 7 numeric digits
  
### Date of birth

  - must be present
  - the teacher must be between 18 and 100 years old
```

claire-hughez (commented):
Looks good Joe, left some thoughts :)

documentation/service-rules/appropriate-body-users.md
claire-hughez: I've pinged @peteryates on this one as I know this is based on his original bit

documentation/service-rules/appropriate-body-users.md
joe-harrison-dfe: Yeah I included them as they are the two options for searching for an ECT if you don't just scroll through the list. Not sure if they belong here though?

documentation/service-rules/appropriate-body-users.md
peteryates: I wonder if maybe it's worth having a separate rule for viewing an ECT which lists the visible fields. Then we don't need to include them here where they're not necessarily relevant to the act of releasing?

documentation/service-rules/appropriate-body-users.md
peteryates: I don't mind the nested list for validated fields/rules, but I'm easy. Maybe leave this one for now so they're consistent and we can change them all to the preferred format later?

documentation/service-rules/appropriate-body-users.md
claire-hughez: Emily will know better than me, but I think it makes more sense not to here, as it's not central to the release journey I believe.

documentation/service-rules/appropriate-body-users.md
claire-hughez: Sure, I know you love your nests 🪺

PR #868, opened 2024-12-03
Rename `object_id` to `period_id`

peteryates:
object_id is a [method on all ruby objects](https://ruby-doc.org/core-2.4.0/Object.html#method-i-object_id) and this clash causes warnings to appear in the deployed environments.


tonyheadford (approved):
Thanks, been meaning to change that 🥇

PR #869, opened 2024-12-03
[765]: signpost user when mentor doesn't have a trn

ltello:
[Issue](https://github.com/DFE-Digital/register-early-career-teachers/issues/765)

https://github.com/user-attachments/assets/9352e826-fbf1-4213-a6fd-5285984d8dac


app/wizards/schools/register_mentor/cannot_without_trn_step.rb
```diff
@@ -0,0 +1,6 @@
+module Schools
+  module RegisterMentor
+    class CannotWithoutTRNStep < Step
```
TobyRet, line 3: Cannot (do what?) without a TRN? 
Register?

Could be worth renaming to make this clear

app/wizards/schools/register_mentor/wizard.rb
```diff
@@ -6,6 +6,7 @@ class Wizard < DfE::Wizard::Base
       steps do
         [
           {
+            cannot_without_trn: CannotWithoutTRNStep,
```
TobyRet, line 9: https://github.com/DFE-Digital/register-early-career-teachers/pull/869#discussion_r1867702676

app/wizards/schools/register_mentor/cannot_without_trn_step.rb
ltello: Happy to change it but is it not too verbose given that we are inside the `Schools::RegisterMentor` namespace?

ltello: Renamed to NoTRNStep

TobyRet (approved):

Issue #821, opened 2024-11-27
Set up Blazer to help us query data and see our data model


claire-hughez:
## Description

We agreed that we should use Blazer to query our database and view our data model.

This would only be used by our team, by those who were confident using SQL. However, for now we can set this up for any admin users. We need to make sure any version of Blazer with real migrated data could only be accessed by those with an admin account.

We agreed Blazer would work best running on a daily snapshot. This is so any queries wouldn't cause performance issues.

We also agreed Blazer should be read-only. No one should be able to write new or different data to the database with Blazer.

## Why?

Blazer is a tool we can use to see the data and the new shape of our data model easily, even if you're not a developer. We imagine it would be PMs and mainly analysts using it. This wouldn't be for official purposes, e.g. funding or statistics.

Blazer is helpful to have in addition to Big Query (which we'll need to add later) because:
1. We can get personal details using it but not Big Query, which may be useful for comms or other purposes
2. Non-devs with enough SQL knowledge can do simple queries for information
3. It's much more user-friendly than Big Query
4. It shows our new data model really clearly

You can [read why we agreed this would be helpful here.](https://github.com/DFE-Digital/register-early-career-teachers/issues/733)

## User story

**As** a product manager
**I need** to be nosey and see our data
**So that** I can make better decisions and bother developers less

## Prototype, screenshots & content
In the thread [linked here](https://ukgovernmentdfe.slack.com/archives/C06V31R0Z17/p1732701195204989) you can see some screenshots of how it's set up on similar services. I don't think we need anything more, but shout if we do.

## Acceptance criteria

- [ ] Create a way for us to query data using Blazer
- [ ] Make sure it has read-only access
- [ ] Make sure it can only be accessed by those with admin accounts, not appropriate body accounts or personas
- [ ] Make sure it uses a daily snapshot for data

## Technical notes

- [ ] Use the path `/admin/blazer`
- [ ] We'll need a secondary snapshot database and a scheduled task that backs up our production database and restores it to the snapshot
- [ ] Blazer should connect to the snapshot database only, so will need the appropriate config in the keyvault.

gracecbrennan:
This isn't too far away

gracecbrennan:
Pete and Toby going to try to chat on this today

gracecbrennan:
final check from Pete needed on this

PR #884, opened 2024-12-04
Set up Blazer 

TobyRet:
### Ticket 
#821 

### Changes
- Add Blazer and configuration

### Notes
- Will not be available on live / prod
- Can be accessed at '/admin/blazer' - you need to sign in as a DFE / admin person to access

blazer_on_review_app


.env.example
```diff
@@ -8,3 +8,4 @@ ENABLE_PERSONAS=true
 ENABLE_MIGRATION_TESTING=true
 ENABLE_SCHOOLS_INTERFACE=true
 TEST_GUIDANCE=true
+ENABLE_BLAZER=true
```
TobyRet, line 11: This is to allow us to restrict which environment Blazer is accessed (I'm assuming we do not want on live / production right now)

config/blazer.yml
```diff
@@ -0,0 +1,79 @@
+# see https://github.com/ankane/blazer for more info
+
+data_sources:
+  main:
+    url: <%= ENV["BLAZER_DATABASE_URL"] %>
```
TobyRet, line 5: Do we have a url for the read-only db?

app/controllers/application_controller.rb
```diff
@@ -8,6 +8,12 @@ class ApplicationController < ActionController::Base
 
 private
 
+  def require_persona_admin
```
TobyRet, line 11: This seems a bit pointless whilst we are using personas? Should we set separate Basic Auth credentials instead?
[27bc33a](https://github.com/ankane/blazer?tab=readme-ov-file#basic-authentication)

app/controllers/application_controller.rb
```diff
@@ -8,6 +8,11 @@ class ApplicationController < ActionController::Base
 
 private
 
+  def require_admin
+    # This method is used by Blazer to restrict access. See config/blazer.yml
+    redirect_to('/sign-in') unless Admin::Access.new(current_user).can_access?
```
TobyRet, line 13: @tonyheadford - can you confirm that this check for admin users will suffice for the foreseeable?

tonyheadford: Yes I think so

tonyheadford (approved):

Issue #769, opened 2024-11-21
Report mentor email address, CYA and submit


mason-emily:
## Description
This story delivers further steps in the journey for assigning and registering a new mentor for an ECT, including entering the mentor's email, checking your answers and submitting to reach the confirmation page.

## User story

**As** school user
**I need** to report the mentor's email address and check the details I've entered before submitting
**So that** DfE have the correct details they need to enable mentoring and/or mentor training

## Prototype, screenshots & content (p: ecf2)
[Prototype email address page](https://ecf2-prototype-e3c7aba62d82.herokuapp.com/v5/school/mentor/email-address)
[Prototype CYA & confirmation page](https://ecf2-prototype-e3c7aba62d82.herokuapp.com/v5/school/mentor/check-answers)

Note, their are additional pages in the prototype for collecting training details which are **not** in scope for this ticket.

**Screenshots of relevant screens:**

![image](https://github.com/user-attachments/assets/aa3ef8fb-5c16-497b-8ba3-b4fb32c49e1f)

Content should be as per the prototype above.
URL and validation messages should be as per [this doc](https://educationgovuk.sharepoint.com/:x:/r/sites/TeacherServices/Shared%20Documents/Teacher%20Continuing%20Professional%20Development/Teacher%20CPD%20Team/Register%20early%20career%20teachers/Beta/Design/Register%20ECTs%20page%20titles%20and%20URLs.xlsx?d=w0c1578ea643147b29357c7f610ee178f&csf=1&web=1&e=U8iIdi).

## Acceptance criteria

- [ ] 1. When user clicks 'Confirm and continue' on the <ins>'Check mentor details'</ins> page (#759), they will be navigated to the <ins>'What is [mentor name]'s email address?'</ins> page.
- [ ] 2. On the <ins>'What is [mentor name]'s email address?'</ins> page:
    - [ ] 2a. The H1 question will pull in the mentor's name e.g. What is [Samuel Taylor]'s email address?
[...]

PR #886, opened 2024-12-04
[#769]: report mentor email address cya and submit

ltello:
[Issue](https://github.com/DFE-Digital/register-early-career-teachers/issues/769)

Better review commit by commit.

Video: 

https://github.com/user-attachments/assets/d3c7275f-0db3-49ce-aabf-842e5c15d8c5


lib/session_repository.rb
```diff
@@ -22,8 +23,11 @@ def get(key)
     store[key.to_s]
   end
 
-  def method_missing(name, *_args)
-    get(name)
+  # Every missing methodname ending with '=' will call #update(name: args)
+  # Otherwise call #get(name)
+  def method_missing(name, *args)
```
TobyRet, line 28: Given the recent chats would be good to here from other @peteryates @tonyheadford  if they have a view on using `method_missing`. If we do keep it, has it now reached a level of complexity where we need to have unit tests for this?

app/controllers/schools/register_mentor_controller.rb
```diff
@@ -7,7 +7,7 @@ class RegisterMentorController < SchoolsController
     WIZARD_CLASS = Schools::RegisterMentor::Wizard.freeze
 
     def start
-      @ect_name = params[:ect_name]
+      @ect_name = session[:register_mentor_for_ect_named] = params[:ect_name]
```
TobyRet, line 10: Chaining assignments in this way can make it more difficult to read / debug?
Would the following be acceptable?
```
@ect_name = params[:ect_name]
session[:register_mentor_for_ect_named] = @ect_name
```

app/wizards/schools/register_mentor/mentor.rb
```diff
@@ -1,3 +1,9 @@
+# Steps in the Schools::RegisterMentor wizard will have a Schools::RegisterMentor::Mentor instance
```
TobyRet, line 1: The pattern we decide to use here should also be applied to Register ECT. 
Might be best if @peteryates @tonyheadford  look at this with fresh eyes (you already know my opinion on this ;-) )
Will happily go with the consensus

ltello: Sorry but I don't know what pattern you mean

lib/session_repository.rb
ltello: This has been in place for 2-3 weeks. 
Why reviewing it now?

ltello: The idea of the SessionRepository is to create a very very simple storage service for the journey based in a hash stored in the Rails session. 

So we just expose a tiny public interface to get and set fields. 
As we will have many key value pairs stored it's really tedious, verbose and difficult to maintain a getter and a setter method for each of field. Hence the bit of metaprogramming that should do the job for us. I now that metaprogramming has its risks and drawback but this is a very controlled and isolated one.

app/controllers/schools/register_mentor_controller.rb
ltello: Sorry to disagree. It's just a nice, handy feature provided by Ruby for simplicity that I don't think will hurt much.

peteryates: I'd agree with Toby on this, it does take a little more effort to see what's going on.

lib/session_repository.rb
peteryates: I do think this has the potential to making finding bugs more complex. As it's just in one class happy to see it stay for now for the sake of merging, but I think there might be scope for doing this more simply with an openstruct.

peteryates (approved):
Looks good. Left a couple of small comments to think about but would rather see this merged and iterated as it's already a big PR.

PR #888, opened 2024-12-04
Minor amends to content strategy

robthompson16:
Minor amends to entries for appropriate body, delivery partner and lead provider to add text about what they're responsible for.

Fix bullet list

### DIFF
--- a/documentation/content-strategy.md
+++ b/documentation/content-strategy.md
@@ -3,8 +3,6 @@
 ## Style guide
 
 This style guide covers style points for content on the Register early career teachers service. It's been designed to be used with the [GOV.UK style guide](https://www.gov.uk/guidance/style-guide) and [DfE style guide](https://design.education.gov.uk/design-system/style-guide).
-
-**appoint an appropriate body** - a school appoints an appropriate body to quality assure an ECT's induction [note: appoint likely to be replaced with nominate in the future] 
  
 **appropriate body** - responsible for assuring the quality of the statutory induction of ECTs, lower case, don't abbreviate, explain why we're asking for this information at the point of collecting it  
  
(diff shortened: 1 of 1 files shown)


documentation/content-strategy.md
```diff
@@ -6,13 +6,13 @@ This style guide covers style points for content on the Register early career te
 
 **appoint an appropriate body** - a school appoints an appropriate body to quality assure an ECT's induction [note: appoint likely to be replaced with nominate in the future] 
  
-**appropriate body** - lower case, don't abbreviate, explain why we're asking for this information at the point of collecting it  
+**appropriate body** - responsible for assuring the quality of the statutory induction of ECTs, lower case, don't abbreviate, explain why we're asking for this information at the point of collecting it
```
peteryates, line 9: Maybe worth saying here that appropriate bodies claim ECTs too, which will be the case for the foreseeable future.

robthompson16: @peteryates discussed this at SU and Claire and I agreed that on balance it would be better not to include the term claim here as unlikely to be understood by SITs.

Is it ok to go ahead and merge?

peteryates: Sorry I think I left my comment on the wrong line, I was referring to this bit:

> a school appoints an appropriate body to quality assure an ECT's induction

For the first year of the service's life this won't be how it works. But Yeah, feel free to merge.

peteryates (approved):

documentation/content-strategy.md
robthompson16: We seemed to have avoided using this language in ECF2 anyway by simply asking 'Which appropriate body will be supporting [ECT name]'s induction?

I wonder whether we remove that line altogether? I'm not sure it's necessary to have both lines anyway

robthompson16: As per my comment above I've removed the line about which appropriate body

PR #898, opened 2024-12-05
Update TRN validation

TobyRet:
### Ticket
https://github.com/DFE-Digital/register-early-career-teachers/issues/835 

### Changes
We had already lifted the TRN validator from ECF1 however we were not allowing users to enter old-style TRN numbers in the Register ECT flow. This PR makes the following changes:

- Change the TRN field in the form into a text field so that users can enter old style numbers
- Pass the 'formatted' TRN to the TRS API

### DIFF
--- a/app/views/schools/register_ect/find_ect.html.erb
+++ b/app/views/schools/register_ect/find_ect.html.erb
@@ -5,7 +5,7 @@
 <%= form_with(model: @wizard.current_step, url: @wizard.current_step_path) do |f| %>
   <%= content_for(:error_summary) { f.govuk_error_summary } %>
 
-  <%= f.govuk_number_field(
+  <%= f.govuk_text_field(
         :trn,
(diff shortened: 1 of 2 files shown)


app/views/schools/register_ect/find_ect.html.erb
```diff
@@ -5,7 +5,7 @@
 <%= form_with(model: @wizard.current_step, url: @wizard.current_step_path) do |f| %>
   <%= content_for(:error_summary) { f.govuk_error_summary } %>
 
-  <%= f.govuk_number_field(
+  <%= f.govuk_text_field(
         :trn,
         label: { text: "Teacher reference number (TRN)", size: "m" },
         hint: { text: "Must be 7 digits long", size: "s" },
```
ltello, line 11: Maybe we should align the hint text with the new valid input from the user?

ltello (approved):

PR #912, opened 2024-12-08
Session user

peteryates:
Sorry, this is a bigger PR than intended because it affected lots of tests.

It aims to consolidate the three kinds of users we have:

* DfE Sign-in users
* Persona users
* Admin users (OTP)

into a single `Sessions::SessionUser` class which we can interact with safely regardless of user type. This is needed because the types aren't consistent with each other. For example, OTP users are linked with `User` records and have an `id`, but DfE Sign-in users won't.

Adding this class broke lots of tests. It makes sense to review this commit-by-commit.


app/assets/javascripts/autocomplete.js
```diff
@@ -1,266 +0,0 @@
-import accessibleAutocomplete from 'accessible-autocomplete'
```
TobyRet, line 1: Should this file have been deleted? If so how does this relate to the PR?

app/services/sessions/session_manager.rb
```diff
@@ -4,50 +4,37 @@ class SessionManager
...
       session['user_session'] = {
         'email' => user_info.info.email,
         'name' => user_info.info.then { |info| "#{info.first_name} #{info.last_name}" },
         'organisation_id' => user_info.extra.raw_info.organisation.id,
         'provider' => user_info.provider,
         'urn' => user_info.extra.raw_info.organisation.urn,
-        **session_defaults
+        'last_active_at' => Time.zone.now
       }
     end
```
tonyheadford, line 30: Why doesn't this method create the hash via `SessionUser`?

peteryates: Well spotted, it did for a bit and I must have reverted it! Re-done in the latest commit.

tonyheadford (approved):
Look great to me, would be good if another dev could also review as I've been involved in this

PR #918, opened 2024-12-09
Adding more info for viewing failures

tonyheadford:
Adding more detail to teacher views
 - Error tag for teachers with migration errors in the main list view
 - Warning text to show that not all the teacher's records could be migrated in the teacher view
 - Summary of latest info in teacher view
 - Links to mentors/mentees and to School in teacher view
 - View of original `ParticipantProfile` and `InductionRecord`s for failures

Apologies for lack of screen shots but these would have prod data in and involve much redacting before showing here.

### DIFF
--- a/app/migration/induction_record_sanitizer.rb
+++ b/app/migration/induction_record_sanitizer.rb
@@ -1,7 +1,12 @@
 class InductionRecordSanitizer
   include Enumerable
 
-  class InductionRecordError < StandardError; end
+  class InductionRecordError < StandardError
+    def initialize(participant_profile_id)
+      msg = "#{self.class.name}|#{participant_profile_id}"
+      super(msg)
+    end
+  end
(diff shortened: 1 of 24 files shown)


claire-hughez:
@TobyRet to have a look

app/migration/induction_record_sanitizer.rb
```diff
@@ -57,26 +63,26 @@ def different?(ir1, ir2)
   end
 
   def has_induction_records!
```
TobyRet, line 65: Curious to know why these methods and the ones below use '!' as opposed to '?' given we are checking whether a certain condition exists (rather than performing a 'dangerous' action)

tonyheadford: Yes, they are called to check for the existence of the things etc and raise an error if the condition is not met

Issue #834, opened 2024-11-28
Assign a mentor from the end of the ECT registration journey


mason-emily:
## Description
This story updates the existing Confirmation page in the Add ECT journey (#478) with iterated content and a button that takes the user to the journey for assigning a mentor. 

## User story

**As** school user
**I need** to know what to do next after adding an ECT
**So that** I can report all the details required to DfE

## Prototype, screenshots & content
[Link to prototype](https://ecf2-prototype-e3c7aba62d82.herokuapp.com/)

Screenshots of relevant screens:

Content should be as per the screenshots above.
URL and validation messages should be as per [this doc](https://educationgovuk.sharepoint.com/:x:/r/sites/TeacherServices/Shared%20Documents/Teacher%20Continuing%20Professional%20Development/Teacher%20CPD%20Team/Register%20early%20career%20teachers/Beta/Design/Register%20ECTs%20page%20titles%20and%20URLs.xlsx?d=w0c1578ea643147b29357c7f610ee178f&csf=1&web=1&e=U8iIdi).

## Acceptance criteria

- [ ] 1. The content and design of the Confirmation page in the Add ECT journey (#478) will be updated as per the prototype (green box removed and updated content)
- [ ] 2. There will be a new button to 'Assign a mentor', which navigates the user to the start of the journey for assigning a mentor to that ECT. Note, if there are existing mentors currently active at the school, the start of the mentor journey that the user will be navigated to will be the <ins>'Who will mentor [ECT name]?'</ins> page (#777). If there are no existing mentors, the start of the mentor journey is the <ins>'What you'll need to add a new mentor for [ECT name]'</ins> page (#758).
**- [ ] 3. BACK LINKS???**

## Linked tickets
#478 #777 #758

claire-hughez:
@JSJohal are you ok to confirm on back links please?

JSJohal:
@claire-hughez have updated the acceptance criteria :)

robthompson16:
Screen shot added @claire-hughez

claire-hughez:
thank you! <3 @robthompson16

claire-hughez:
PR will be up after standup

JSJohal:
Ready for product review @claire-hughez

PR #929, opened 2024-12-10
Assign a mentor from the end of the ECT registration journey

TobyRet:
### Ticket
#834 

### Change 
- Update the confirmation step in the Register ECT wizard to allow users to assign a mentor

assign_a_mentor

### DIFF
--- a/app/views/schools/register_ect_wizard/confirmation.html.erb
+++ b/app/views/schools/register_ect_wizard/confirmation.html.erb
@@ -1,5 +1,13 @@
-<%= govuk_panel(title_text: "You have registered #{@ect.full_name} as an early career teacher") %>
+<% page_data(title: "You have saved #{@ect.full_name}'s details") %>
 
-<h2 class="govuk-heading-m">What happens next</h2>
-<p class="govuk-body">We’ll let this person know you’ve registered them for ECF-based training at your school. They do not need to provide us with any more information.</p>
-<%= govuk_link_to "Back to your ECTs", schools_ects_home_path %>
+<p class="govuk-body">We'll email them to let them know. They do not need to provide us with any more information.</p>
+
+<h2 class="govuk-heading-m">You still need to assign a mentor</h2>
+<p class="govuk-body">You can either assign a mentor now or return and do it later. You need to do this to complete
+  their registration.</p>
+
+<%= govuk_button_link_to("Assign a mentor", schools_register_mentor_wizard_start_path) %>
+
+<p class="govuk-body">
+  <%= govuk_link_to "Back to your ECTs", schools_ects_home_path %>
+</p>
--- a/app/wizards/schools/register_ect_wizard/ect.rb
+++ b/app/wizards/schools/register_ect_wizard/ect.rb
@@ -3,7 +3,7 @@ module Schools
     class ECT < SimpleDelegator
       # This class is a decorator for the SessionRepository
       def full_name
-        corrected_name || [trs_first_name, trs_last_name].join(" ").strip
+        (corrected_name.presence || [trs_first_name, trs_last_name].join(" ").strip)
       end
 
       def govuk_date_of_birth
--- a/spec/views/schools/register_ect_wizard/confirmation.html.erb_spec.rb
+++ b/spec/views/schools/register_ect_wizard/confirmation.html.erb_spec.rb
@@ -1,0 +1,28 @@
+RSpec.describe "schools/register_ect_wizard/confirmation.html.erb" do
+  let(:ect) { double('ECT', full_name: 'John Doe') }
+  let(:title) { "You have saved #{ect.full_name}'s details" }
+  let(:ects_link) { schools_ects_home_path }
+  let(:assign_mentor_link) { schools_register_mentor_wizard_start_path }
+
+  before do
+    assign(:ect, ect)
+  end
+
+  it "sets the page title to 'You have saved John Doe's details" do
+    render
+
+    expect(sanitize(view.content_for(:page_title))).to eql(sanitize(title))
+  end
+
+  it "includes a link to view all ECTs" do
+    render
+
+    expect(rendered).to have_link('Back to your ECTs', href: ects_link)
+  end
+
+  it "includes a link to assign a mentor" do
+    render
+
+    expect(rendered).to have_link('Assign a mentor', href: assign_mentor_link)
+  end
+end


app/views/schools/register_ect_wizard/confirmation.html.erb
```diff
@@ -1,5 +1,13 @@
...
-<h2 class="govuk-heading-m">What happens next</h2>
-<p class="govuk-body">We’ll let this person know you’ve registered them for ECF-based training at your school. They do not need to provide us with any more information.</p>
-<%= govuk_link_to "Back to your ECTs", schools_ects_home_path %>
+<p class="govuk-body">We'll email them to let them know. They do not need to provide us with any more information.</p>
+
+<h2 class="govuk-heading-m">You still need to assign a mentor</h2>
+<p class="govuk-body">You can either assign a mentor now or return and do it later. You need to do this to complete
+  their registration.</p>
+
+<%= govuk_button_link_to("Assign a mentor", schools_register_mentor_wizard_start_path(ect_id: @ect.ect_at_school_period_id)) %>
```
ltello, line 9: I think this is required to link to: `assign_or_create_mentor_path(@ect.ect_at_school_period)` instead?

app/wizards/schools/register_ect_wizard/ect.rb
```diff
@@ -3,7 +3,7 @@ module RegisterECTWizard
     class ECT < SimpleDelegator
       # This class is a decorator for the SessionRepository
       def full_name
-        corrected_name || [trs_first_name, trs_last_name].join(" ").strip
+        (corrected_name.presence || [trs_first_name, trs_last_name].join(" ").strip)
```
ltello, line 6: Can we remove the wrapping `()`?

app/wizards/schools/register_ect_wizard/ect.rb
```diff
@@ -36,6 +36,11 @@ def register!
                                  school_urn:)
                                .register_teacher!
       end
+
```
ltello, line 39: Wouldn't be easier to keep the ECTAtSchoolPeriod returned by the method above and use it later on in the view?

```
attr_reader :ect_at_school_period

def register!
   @ect_at_school_period = Schools::RegisterECT.new(first_name: trs_first_name,
                                                    last_name: trs_last_name,
                                                    corrected_name:,
                                                    trn:,
                                                    school_urn:)
                                               .register_teacher!
end
```

app/wizards/schools/register_ect_wizard/ect.rb
```diff
@@ -36,6 +36,11 @@ def register!
...
+      def ect_at_school_period_id
+        Schools::Teacher.new(school_urn)
+                        .fetch_ect_at_school_period_id(trn: trn)
+      end
     end
```
ltello, line 44: See comment above

app/services/schools/teacher.rb
```diff
@@ -19,9 +19,15 @@ def fetch_etcs_and_mentors
       end
     end
 
+    def fetch_ect_at_school_period_id(trn:)
+      ECTAtSchoolPeriod
+        .joins(:teacher, :school)
+        .find_by(teacher: { trn: trn }, school: { urn: school_urn }).id
+    end
```
ltello, line 26: See my comment below. I don't think we need this?

spec/views/schools/register_ect_wizard/confirmation.html.erb_spec.rb
```diff
@@ -0,0 +1,28 @@
+RSpec.describe "schools/register_ect_wizard/confirmation.html.erb" do
+  let(:ect) { double('ECT', full_name: 'John Doe', ect_at_school_period_id: 22) }
+  let(:title) { "You have saved #{ect.full_name}'s details" }
+  let(:ects_link) { schools_ects_home_path }
+  let(:assign_mentor_link) { schools_register_mentor_wizard_start_path(ect_id: ect.ect_at_school_period_id) }
```
ltello, line 5: See comment above

spec/wizards/schools/register_ect/ect_spec.rb
```diff
@@ -127,4 +128,13 @@
...
     end
   end
+
+  describe('#ect_at_school_period_id') do
+    it 'returns the id of the ECT at school period' do
+      teacher = FactoryBot.create(:teacher, trn:)
+      ect_at_school_period = FactoryBot.create(:ect_at_school_period, teacher: teacher, school: school)
+
+      expect(ect.ect_at_school_period_id).to eq(ect_at_school_period.id)
+    end
```
ltello, line 138: See comment above

ltello (changes requested):

app/wizards/schools/register_ect_wizard/ect.rb
ltello: or maybe just:

```
def school
  @school ||= School_find_by_urn(school_urn)
end

def teacher
  @teacher ||= Teacher.find_by(trn:)
end

def ect_at_school_period
 @ect_at_school_period ||= teacher.ect_at_school_periods.where(school:).last if teacher
end
``` 
and do not change #register!

ltello: or maybe persist the `ect_at_school_period_id` in the `....CheckAnswersStep`:

`app/wizards/schools/register_ect_wizard/check_answers_step.rb`:
```
def persist
  update!(ect_at_school_period_id: ect.register!.id)
end
```

and in this class:

```
def ect_at_school_period
 @ect_at_school_period ||= ECTAtSchoolPeriod.find_by_id(ect_at_school_period_id)
end
```

TobyRet: I think the issue I have with this is that the `register!` method is now doing three things: adding the ECT to the database (writing), fetching the `ect_at_school_period_id` and updating the session.

I think this is doing too much / introducing side effects  - I would prefer to keep writing to the db as a separate / distinct operation. 

Potentially the `confirm_step` could be updated to store the `ect_at_school_period_id` as a separate command ie 

```
 def persist
        ect.register!
        # Something like ... 
        ect.store_ect_at_school_period_id
  end
```   

But I'm not sure if this makes this makes for a better solution. I was thinking also that upon successfully writing to the database we should be destroying the session data anyway?

ltello: In the `persist` method of other steps we are calling `ect.update` like the last option proposed above.

ltello (approved):

PR #935, opened 2024-12-11
Make persona logins a bit more robust

peteryates:
It's possible to get yourself into a bit of a state when using personas, especially if you flit between them.

This change makes that work a bit better by ensuring that when we update a session we reload it so the current session_user is in sync with what's actually in the session['session_user'] hash.

https://github.com/user-attachments/assets/fea61513-1a1f-448c-82a4-936728d1f334


app/controllers/sessions_controller.rb
```diff
@@ -43,9 +43,13 @@ def begin_persona_session(user_info)
       appropriate_body_id: params["appropriate_body_id"].presence,
       school_urn: params["school_urn"].presence
     )
+
+    reload_current_user
```
ltello, line 47: Wouldn't work just by nullifying `@current_user`  in line 23?

I think that way we don't need to redefine a reload current user method and call it.

Also in lines 15, 19 and 33 in `app/services/sessions/session_manager.rb`

